Issue #2, opened 2018-01-02
runtime: Node.js & NPM compatible


yorkie:
### ES6 features

- [x] Promise
- [x] TypedArray
- [x] Arrow Function
- [x] Template String
- [ ] `const/let`

### Built-in Node.js modules

- [x] http
- [x] https
- [x] path
- [x] url
- [x] crypto
  - [x] createHash()
  - [ ] createHmac()
  - [x] randomBytes()
- [x] dns
- [x] events
- [x] process
  - [x] umask()
  - [ ] title
- [x] child_process
  - [x] spawn
  - [x] fork
  - [x] exec
- [x] os
- [x] net
- [x] tls
- [x] file
  - [x] fs.createReadStream
  - [x] fs.createWriteStream
- [x] string_decoder
- [x] zlib
  - [x] streams
  - [ ] sync/async convenient methods
  - [x] constants

### Modules from ecosystem

- [x] debug
- [x] connect
  - [x] connect-router
  - [x] depd
  - [x] body-parser
- [x] iconv-lite
- [x] urllib
- [x] node-fetch
  - [ ] option.compress
- [x] mqtt
  - [x] mqtts
- [x] dbus
  - [x] createService
  - [x] getInterface
  - [x] signal
- [x] lowdb
- [x] mz
- [x] sax

yorkie:
Supported the D-Bus module at 5d94ab828023524f89e2b0dcf21b373d9acdc036 and f662aebce6b6414b46b58a2c74f20eb15045dd9d

yorkie:
Supported the Node.js built-in `os` module at 1b2d2e633cbc19c7abf81becd5a00ff5fbcda55e.

yorkie:
Added `Promise` at 6458f29a1c8c6e948c4a078bcd10c17622375b48.

yorkie:
Updated the TODO list.

yorkie:
The roadmap has been outdated, let's close it :)

PR #6, opened 2018-01-16
dbus: add signal support

yorkie:


yorkie:
Try to complete the dbus-signal part in #2

PR #7, opened 2018-01-17
crypto: init module

yorkie:
Adds the first function `randomBytes()`, currently the asynchronous `randomBytes` is mocked by callback.

BTW, this patch opens the `mbedtls` to be compiled by default :)


yorkie:
This tries to complete #2.

PR #10, opened 2018-01-18
jerry: support the ecma standard oct literal

yorkie:
We should support the oct literal in standard way: https://www.ecma-international.org/ecma-262/8.0/index.html#prod-OctalIntegerLiteral, which is:

```
0o666
0O666
```

### DIFF
--- a/deps/jerry/jerry-core/parser/js/js-lexer.c
+++ b/deps/jerry/jerry-core/parser/js/js-lexer.c
@@ -880,7 +880,8 @@ lexer_parse_number (parser_context_t *context_p) /**< context */
         source_p++;
       }
 
-      do {
+      do 
+      {
         source_p++;
       }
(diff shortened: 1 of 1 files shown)


deps/jerry/jerry-core/parser/js/js-lexer.c
```diff
@@ -873,7 +874,13 @@ lexer_parse_number (parser_context_t *context_p) /**< context */
         parser_raise_error (context_p, PARSER_ERR_OCTAL_NUMBER_NOT_ALLOWED);
```
algebrait: In strict mode, only legacyOctalIntegerLiteral is disallowed, but OctalIntegerLiteral is allowed.

yorkie:
Fixed @algebrait :)

PR #15, opened 2018-01-22
buffer: fix toString and constructor to consistent with node

yorkie:
The first argument of `Buffer.prototype.toString()` should be hex string.


yorkie:
This supported the iconv-lite to work normally :)

PR #20, opened 2018-01-22
http, https: add Agent and os.release() to support urllib

yorkie:
Support urllib for HTTP requests(#2).


yorkie:
Don't use Agent currently :)

PR #37, opened 2018-01-25
es2015: enable some features

yorkie:
The enabled features as follow:

- arrow function
- template string
- `TypedArray`
- `Promise`


yorkie:
Use `--disable-es2015` to disable these features :)

yorkie:
Updates to #2.

PR #60, opened 2018-02-27
Add support for `console.time`

legendecas:


yorkie:
@legendecas Awesome, could you please add some test case for your code, see https://github.com/Rokid/ShadowNode/tree/master/test/run_pass :)

src/js/iotjs.js
```diff
@@ -222,6 +222,32 @@
...
 
+  var _hrtime = process.hrtime;
+  var NANOSECOND_PER_SECONDS = 1e9;
+  process.hrtime = function hrtime (time) {
+    var curr = _hrtime();
+    if (time == null) {
+      return curr;
+    }
+    if (!Array.isArray(time)) {
+      var error = TypeError('[ERR_INVALID_ARG_TYPE]: The "time" argument must be of type Array. Received type ' + typeof time);
```
yorkie: missing the `new` keyword :)

src/js/iotjs.js
```diff
@@ -222,6 +222,32 @@
...
+      error.code = 'ERR_INVALID_ARG_TYPE';
+      throw error;
+    }
+    if (time.length != 2) {
+      var error = TypeError('[ERR_INVALID_ARRAY_LENGTH]: The array "time" (length ' + String(time.length) + ') must be of length 2.');
```
yorkie: Ditto, and the line should be in 80 characters :)

src/js/console.js
```diff
@@ -18,7 +18,23 @@ var util = require('util');
 
 
 function Console() {
-  // Empty
+  var prop = {
+    writable: true,
+    enumerable: false,
+    configurable: true
+  };
+  prop.value = stdout;
```
yorkie: Don't use the same property object for different defines, use `Object.assign()` to generate separate property :)

src/js/console.js
```diff
@@ -40,15 +56,25 @@ Console.prototype.error = function() {
...
-  log: console.log.bind(console),
-  debug: console.debug.bind(console),
-  info: console.info.bind(console),
-  warn: console.warn.bind(console),
-  error: console.error.bind(console),
-  Console: Console,
-  _stdout: stdout,
-  _stderr: stderr,
+  var time = this._times[label];
+  if (time == null) {
```
yorkie: strict equal?

legendecas: `variable == null` is both true if variable is `null` or `undefined`. I think this shall be one exception for rule of strict equal.

yorkie: See https://github.com/airbnb/javascript#comparison--eqeqeq, we should follow this lint rule.

yorkie:
Rad @legendecas !

PR #66, opened 2018-03-01
tcp: throw error instead of raw errno on listen

### LINKED ISSUE #57: net: listen on exists port not thrown error

legendecas:
close #57


src/modules/iotjs_module_tcp.c
```diff
@@ -400,6 +401,12 @@ JS_FUNCTION(Listen) {
   int err = uv_listen((uv_stream_t*)(iotjs_tcpwrap_tcp_handle(tcp_wrap)),
                       backlog, OnConnection);
 
+  if (err < 0) {
+    jerry_value_t jerror = iotjs_create_uv_exception(err, "listen");
+    jerry_value_set_error_flag(&jerror);
+    return jerror;
+  }
```
yorkie: See https://github.com/Rokid/ShadowNode/blob/00a011f48b22ee3f0efae8460f01313a6db1f97f/src/js/net.js#L483 
removing the changes and using `process._createUVException()`

src/js/net.js
```diff
@@ -610,11 +610,17 @@ Server.prototype.listen = function() {
...
+      self._handle.close();
+      self.emit('error', err);
+      return self;
+    }
+  } catch (err) {
     self._handle.close();
-    return self.emit('error', err);
+    self.emit('error', err);
+    return self;
   }
```
yorkie: After using `process._createUVException()`, you don't need `try...catch` to handle this error :)

src/modules/iotjs_module_tcp.c
legendecas: As the error shall be considered as a uv call exception, I think it's better to create the js error object right after the uv calls.
Maybe the native module could return the error object rather than throwing it?

src/modules/iotjs_module_tcp.c
```diff
@@ -400,6 +401,11 @@ JS_FUNCTION(Listen) {
   int err = uv_listen((uv_stream_t*)(iotjs_tcpwrap_tcp_handle(tcp_wrap)),
                       backlog, OnConnection);
 
+  if (err < 0) {
+    jerry_value_t jerror = iotjs_create_uv_exception(err, "listen");
+    return jerror;
+  }
+
```
yorkie: There is no need to make changes to this file.

src/js/net.js
```diff
@@ -611,10 +611,10 @@ Server.prototype.listen = function() {
   self._handle.owner = self;
 
   var err = self._handle.listen(backlog);
-
   if (err) {
     self._handle.close();
-    return self.emit('error', err);
+    self.emit('error', err);
+    return self;
```
yorkie: How about the following:

```js
var err = self._handle.listen(backlog);
if (err < 0) {
  var e = process._createUVException(err, 'net.listen');
  self._handle.close();
  self.emit('error', e);
  return self;
}
```

test/node/parallel/test-net-bind-twice.js
```diff
@@ -46,15 +46,13 @@ server1.listen(0, '127.0.0.1', common.mustCall(function() {
...
 
   server2.on('error', common.mustCall(function(e) {
-    // EADDRINUSE have different value on OSX.
-    if (common.isOSX) {
-      assert.strictEqual(e, -48);
-    } else {
-      assert.strictEqual(e, -98);
-    }
+    assert(e !== null);
+    assert(typeof e === 'object');
```
yorkie: Verify the `e` should be an `Error` object.

yorkie:
Thanks, merged 👍

PR #81, opened 2018-03-19
tls: bug fix

lolBig:
- limit buffer size with 16k(defined by MBEDTLS_SSL_MAX_CONTENT_LEN) for iotjs_tlswrap_encode_data
- tls encode incoming data multi times when data is larger than 16k
- alloc 1m heap memroy for MqttGetPublish buffer instead of dynamic stack memory

### DIFF
--- a/src/js/mqtt.js
+++ b/src/js/mqtt.js
@@ -151,14 +151,20 @@ MqttClient.prototype._keepAlive = function() {
  * @param {Function} callback
  */
 MqttClient.prototype.publish = function(topic, payload, options, callback) {
-  var buf = this._handle._getPublish(topic, {
-    id: this._msgId++,
-    qos: (options && options.qos) || 0,
-    dup: (options && options.dup) || false,
-    retain: (options && options.retain) || false,
-    payload: payload || '',
-  });
-  this._write(buf);
+  var buf;
+  try {
+    buf = this._handle._getPublish(topic, {
+      id: this._msgId++,
+      qos: (options && options.qos) || 0,
+      dup: (options && options.dup) || false,
+      retain: (options && options.retain) || false,
+      payload: payload || '',
+    });
+  } catch (err) {
+    callback(err);
+    return;
+  }
+  this._write(buf, callback);
 };
 
 /**
--- a/src/js/mqtt.js
+++ b/src/js/mqtt.js
@@ -151,20 +151,18 @@ MqttClient.prototype._keepAlive = function() {
  * @param {Function} callback
  */
 MqttClient.prototype.publish = function(topic, payload, options, callback) {
-  var buf;
   try {
-    buf = this._handle._getPublish(topic, {
+    var buf = this._handle._getPublish(topic, {
       id: this._msgId++,
       qos: (options && options.qos) || 0,
       dup: (options && options.dup) || false,
       retain: (options && options.retain) || false,
       payload: payload || '',
     });
+    this._write(buf, callback);
   } catch (err) {
     callback(err);
-    return;
   }
-  this._write(buf, callback);
 };
 
 /**
--- a/src/js/tls.js
+++ b/src/js/tls.js
@@ -98,9 +98,13 @@ TLSSocket.prototype.write = function(data, cb) {
     }
     var chunk = data.slice(sourceStart, sourceStart + sourceLength);
     sourceStart += sourceLength;
-    var encodedChunk = this._tls.write(chunk);
-    if (!Buffer.isBuffer(encodedChunk)) {
-      throw new Error('Encryption is not available');
+    var encodedChunk;
+    try {
+      // tls.write may throw error if iotjs_tlswrap_encode_data failure
+      encodedChunk = this._tls.write(chunk);
+    } catch (err) {
+      cb(err);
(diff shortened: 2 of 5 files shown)


src/modules/iotjs_module_tls.c
```diff
@@ -235,9 +235,16 @@ JS_FUNCTION(TlsConstructor) {
 
 jerry_value_t iotjs_tlswrap_encode_data(iotjs_tlswrap_t_impl_t* _this, 
                                         iotjs_bufferwrap_t* inbuf) {
+  /*
+    inbuf length should less equal than MBEDTLS_SSL_MAX_CONTENT_LEN which is 16384 bytes
```
yorkie: Missing the * before the start of the line.

src/js/tls.js
```diff
@@ -84,10 +85,27 @@ TLSSocket.prototype.write = function(data, cb) {
...
+    if (chunkLeft > chunkMaxByteLength) {
+      sourceLength = chunkMaxByteLength;
+    } else {
+      sourceLength = chunkLeft
+    }
+    var chunk = data.slice(sourceStart, sourceStart + sourceLength);
+    sourceStart += sourceLength;
+    var encodedChunk = this._tls.write(chunk);
+    if (!Buffer.isBuffer(encodedChunk)) {
+      throw new Error('Encryption is not available');
```
yorkie: Shall we just break the `while-loop` here, and throws an error outside of the loop :)

src/modules/iotjs_module_mqtt.c
```diff
@@ -153,6 +153,20 @@ JS_FUNCTION(MqttGetConnect) {
   return retbuf;
 }
 
+
+void alloc_payload_buf(unsigned char **buf, int expected_size, int *alloc_size) {
+  static int buf_size = 1 * 1024 * 1024;
+  static unsigned char *buf_ = NULL;
```
yorkie: Why do we here use `static` variable? Stateless function here would be more clearable :)

src/modules/iotjs_module_mqtt.c
```diff
@@ -153,6 +153,20 @@ JS_FUNCTION(MqttGetConnect) {
...
+  if (buf_ == NULL) {
+    buf_ = malloc((size_t)buf_size * sizeof(unsigned char));
```
yorkie: Check if the `malloc` does success, too :)

yorkie: And where do the `buf_` get freed?

src/modules/iotjs_module_mqtt.c
```diff
@@ -180,18 +197,23 @@ JS_FUNCTION(MqttGetPublish) {
                                   top,
                                   (unsigned char*)iotjs_string_data(&msg_payload_str),
                                   msg_size);
-
+  if (len < 0) {
+    ret = JS_CREATE_ERROR(COMMON, "mqtt payload is too large");
+    goto exit;
```
yorkie: Use `if-else` to replace the unnecessary `goto` statements :)

src/js/mqtt.js
```diff
@@ -151,14 +151,20 @@ MqttClient.prototype._keepAlive = function() {
...
+      qos: (options && options.qos) || 0,
+      dup: (options && options.dup) || false,
+      retain: (options && options.retain) || false,
+      payload: payload || '',
+    });
+  } catch (err) {
+    callback(err);
+    return;
+  }
+  this._write(buf, callback);
```
yorkie: Put this line inside try block :)

src/js/mqtt.js
```diff
@@ -151,14 +151,20 @@ MqttClient.prototype._keepAlive = function() {
...
 MqttClient.prototype.publish = function(topic, payload, options, callback) {
-  var buf = this._handle._getPublish(topic, {
-    id: this._msgId++,
-    qos: (options && options.qos) || 0,
-    dup: (options && options.dup) || false,
-    retain: (options && options.retain) || false,
-    payload: payload || '',
-  });
-  this._write(buf);
+  var buf;
```
yorkie: Ditto

src/js/tls.js
```diff
@@ -6,6 +6,7 @@ var net = require('net');
 var util = require('util');
 var EventEmitter = require('events').EventEmitter;
 var TlsWrap = native.TlsWrap;
+var chunkMaxByteLength = 16 * 1024;
```
yorkie: Could we get the value from mbedtls defines?

src/js/tls.js
```diff
@@ -84,10 +85,31 @@ TLSSocket.prototype.write = function(data, cb) {
...
+    if (chunkLeft > chunkMaxByteLength) {
+      sourceLength = chunkMaxByteLength;
+    } else {
+      sourceLength = chunkLeft
+    }
+    var chunk = data.slice(sourceStart, sourceStart + sourceLength);
+    sourceStart += sourceLength;
+    var encodedChunk;
+    try {
+      // tls.write may throw error if iotjs_tlswrap_encode_data failure
```
yorkie: Add the `TODO` or `FIXME` header before every comment :)

PR #88, opened 2018-03-22
mqtt: memory and connection optimization

lolBig:
- implement mqtt reconnect
- optimize memory alloc in mqtt, use heap instead of stack
- fix mqtt unsubscribe crash
- fix memory leak (invoke iotjs_jval_get_property without jerry_release_value in subscribe)
- catch all native invoke


src/js/mqtt.js
```diff
@@ -65,84 +69,138 @@ MqttClient.prototype.connect = function() {
...
+  } catch (err) {
+    this.disconnect(err);
+    return;
+  }
+  this._write(buf);
 };
 
+MqttClient.prototype._onerror = function (err) {
+  this._ondisconnect(err);
+}
```
yorkie: missing ;

src/js/mqtt.js
```diff
@@ -65,84 +69,138 @@ MqttClient.prototype.connect = function() {
...
+
+MqttClient.prototype._onend = function () {
+  this._ondisconnect();
+}
```
yorkie: ditto

src/js/mqtt.js
```diff
@@ -65,84 +69,138 @@ MqttClient.prototype.connect = function() {
...
+MqttClient.prototype._ondisconnect = function (err) {
+  if (err) {
+    this.emit('error', err);
+  }
+  if (this._isConnected) {
+    this._isConnected = false;
+    this.emit('offline');
+  }
+  this.reconnect();
+}
```
yorkie: Ditto!

src/js/mqtt.js
```diff
@@ -65,84 +69,138 @@ MqttClient.prototype.connect = function() {
...
+ * TODO
  * @method _write
  * @param {Buffer} buffer
  * @param {Function} callback
  */
 MqttClient.prototype._write = function(buffer, callback) {
-  this._socket.write(buffer, (err, data) => {
+  callback = callback || noop;
+  if (!this._isConnected) {
+    throw new Error('mqtt is disconnected');
```
yorkie: use `EventEmitter.prototype.emit`

src/js/mqtt.js
```diff
@@ -65,60 +69,87 @@ MqttClient.prototype.connect = function() {
...
+      // send publish ack
+      try {
+        var ack = this._handle._getAck(msg.id, msg.qos);
+        this._write(ack);
+      } catch (err) {
+        this.disconnect(err);
+      }
+    }
+  } else {
+    // TODO handle other message type
```
yorkie: Needs throw some error or log something.

src/js/mqtt.js
```diff
@@ -191,37 +260,38 @@ MqttClient.prototype.subscribe = function(topic, options, callback) {
...
+    topic = [topic];
+  }
+  try {
+    var buf = this._handle._getUnsubscribe(topic, {
+      id: this._msgId++,
     });
+    this._write(buf, callback);
+  } catch (err) {
+    callback(err);
   }
```
yorkie: Why use `try-catch` here?

src/modules/iotjs_module_mqtt.c
```diff
@@ -94,6 +94,22 @@ int iotjs_mqtt_read_buf(unsigned char* buf, int len, void* ctx) {
   return len;
 }
 
+static void iotjs_mqtt_alloc_buf(unsigned char **buf, int expected_size, int *alloc_size) {
+  static int buf_size = 128 * 1024;
+  static unsigned char *buf_ = NULL;
+  if (buf_ == NULL) {
+    buf_ = malloc((size_t)buf_size * sizeof(unsigned char));
+    if (buf_ == NULL) {
+      return;
```
yorkie: We should throw an OOM error to abort the whole process.

lolBig: should we just abort the program or exit with some code?

yorkie: Let's ignore this one for this time.

src/js/mqtt.js
lolBig: i would like to log something here, but this will write a lot of messages to stdout after the program run a long time, is it a good choice?

yorkie: Emit another event `unhandled` would good to you?

lolBig: seems like a good idea

src/js/mqtt.js
lolBig: this._handle._getUnsubscribe may throw an error which means buffer is not enough

yorkie:
Awesome works @lolBig 👍

PR #97, opened 2018-03-29
tls: do handshake in async way

yorkie:
This patch fixes the slow tls connection, the bug is caused by we put the handshake into the main thread, now we spread out into other workers and smoothing the JerryScript thread.

screen shot 2018-03-30 at 6 26 02 am

The workers still in idle even we put the tls handshake there, the next plan is move http parsing works there, too.

/cc @Rokid/node-js


yorkie:
Closed this because it's not necessary :)

PR #108, opened 2018-04-26
tls: break if tls has been destroyed

yorkie:


yorkie:
Updated @lolBig Added the magic number :)

PR #119, opened 2018-05-15
stacktrace: get full stacktrace from vm context chain

legendecas:


src/modules/iotjs_module_process.c
```diff
@@ -207,6 +207,22 @@ JS_FUNCTION(Loadstat) {
 }
 
 
+JS_FUNCTION(GetStackFrames) {
+  // clean dump file
+  jerry_cleanup_parser_dump_file();
```
yorkie: We should not remove it here, it might be used later if something is broken :(

src/modules/iotjs_module_process.c
```diff
@@ -207,6 +207,27 @@ JS_FUNCTION(Loadstat) {
 }
 
 
+JS_FUNCTION(GetStackFrames) {
+  // create frames
+  uint32_t* frames = jerry_get_stacktrace();
+  uint32_t len = jerry_get_stacktrace_max_depth();
```
algebrait: why not trace through context chain like vm.c?

algebrait: it's better to provide api jerry_get_stacktrace_depth

algebrait: seems use stack_index instead of jerry_get_stacktrace_max_depth is ok

algebrait:
LGTM

yorkie:
@legendecas The unit test case is needed as well, see https://github.com/Rokid/ShadowNode/tree/master/test/run_pass :)

yorkie:
Rad works @legendecas thank you :p

PR #122, opened 2018-05-18
stacktrace: attach stack trace to error instance

### LINKED ISSUE #118: attach stack trace to error object

legendecas:
Expected behavior:
```js
function namedFunction () {
  return new Error('foobar')
}

console.log(namedFunction())
// Error: foobar
//     at namedFunction (/some/where/the/script/locates.js:1:10)


function capturePlainObject () {
  var e = {}
  Error.captureStackTrace(e)
  return e
}

console.log(capturePlainObject().stack)
//     at capturePlainObject (/some/where/the/script/locates.js:10:10)
```

close #118

### DIFF
--- a/deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
+++ b/deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
@@ -149,17 +149,27 @@ ecma_new_standard_error (ecma_standard_error_t error_type) /**< native error typ
   // create frames
   uint32_t *frames = malloc(sizeof(uint32_t) * depth);
   memset(frames, 0, sizeof(uint32_t) * depth);
-  jerry_get_stacktrace_depth(frames, depth);
+  jcontext_get_backtrace_depth(frames, depth);
 
-  jerry_value_t jframes = jerry_create_array(depth);
+  ecma_value_t frames_array_length_val = ecma_make_uint32_value (depth);
+  ecma_value_t frames_array = ecma_op_create_array_object (&frames_array_length_val, 1, true);
+  ecma_free_value (frames_array_length_val);
+  ecma_object_t *array_p = ecma_get_object_from_value (frames_array);
+
   for (uint32_t i = 0; i < depth; ++i) {
-    jerry_set_property_by_index(jframes, i, jerry_create_number(frames[i]));
+    ecma_string_t *str_idx_p = ecma_new_ecma_string_from_uint32 ((uint32_t) i);
+    ecma_property_value_t *index_prop_value_p = ecma_create_named_data_property (array_p,
+                                                                           str_idx_p,
+                                                                           ECMA_PROPERTY_CONFIGURABLE_WRITABLE,
+                                                                           NULL);
+    ecma_deref_ecma_string (str_idx_p);
+    index_prop_value_p->value = ecma_make_uint32_value(frames[i]);
   }
 
   free(frames);
 
-  prop_value_p->value = jframes;
-  ecma_deref_object (ecma_get_object_from_value (jframes));
+  prop_value_p->value = frames_array;
+  ecma_deref_object (array_p);
 
   return new_error_obj_p;
 } /* ecma_new_standard_error */
--- a/deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
+++ b/deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
@@ -134,6 +134,32 @@ ecma_new_standard_error (ecma_standard_error_t error_type) /**< native error typ
 
   ((ecma_extended_object_t *) new_error_obj_p)->u.class_prop.class_id = LIT_MAGIC_STRING_ERROR_UL;
 
+  const char *frames_id_p = "__frames__";
+
+  ecma_string_t *frames_str_p = ecma_new_ecma_string_from_utf8 ((const lit_utf8_byte_t *) frames_id_p, 10);
+
+  ecma_property_value_t *prop_value_p = ecma_create_named_data_property (new_error_obj_p,
+                                                                         frames_str_p,
+                                                                         ECMA_PROPERTY_CONFIGURABLE_WRITABLE,
+                                                                         NULL);
+  ecma_deref_ecma_string (frames_str_p);
+
+  uint32_t depth = 10;
+  // create frames
+  uint32_t* frames = malloc(sizeof(uint32_t) * depth);
+  memset(frames, 0, sizeof(uint32_t) * depth);
+  jerry_get_stacktrace_depth(frames, depth);
+
+  jerry_value_t jframes = jerry_create_array(depth);
+  for (uint32_t i = 0; i < depth; ++i) {
+    jerry_set_property_by_index(jframes, i, jerry_create_number(frames[i]));
+  }
+
+  free(frames);
+
+  prop_value_p->value = jframes;
+  ecma_deref_object (ecma_get_object_from_value (jframes));
+
   return new_error_obj_p;
 } /* ecma_new_standard_error */
 
--- a/deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
+++ b/deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
@@ -15,8 +15,8 @@
 
 #include <stdarg.h>
 #include <stdlib.h>
+#include "ecma-array-object.h"
 #include "ecma-builtins.h"
-#include "ecma-builtin-helpers.h"
 #include "ecma-conversion.h"
 #include "ecma-exceptions.h"
 #include "ecma-gc.h"
--- a/deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
+++ b/deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
@@ -14,6 +14,7 @@
  */
 
 #include <stdarg.h>
+#include <stdlib.h>
 #include "ecma-builtins.h"
 #include "ecma-conversion.h"
 #include "ecma-exceptions.h"
--- a/deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
+++ b/deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
@@ -135,9 +135,9 @@ ecma_new_standard_error (ecma_standard_error_t error_type) /**< native error typ
 
   ((ecma_extended_object_t *) new_error_obj_p)->u.class_prop.class_id = LIT_MAGIC_STRING_ERROR_UL;
 
-  const char *frames_id_p = "__frames__";
+  const char *frames_id_p = "_frames";
 
-  ecma_string_t *frames_str_p = ecma_new_ecma_string_from_utf8 ((const lit_utf8_byte_t *) frames_id_p, 10);
+  ecma_string_t *frames_str_p = ecma_new_ecma_string_from_utf8 ((const lit_utf8_byte_t *) frames_id_p, 7);
 
   ecma_property_value_t *prop_value_p = ecma_create_named_data_property (new_error_obj_p,
                                                                          frames_str_p,
@@ -147,7 +147,7 @@ ecma_new_standard_error (ecma_standard_error_t error_type) /**< native error typ
 
   uint32_t depth = 10;
   // create frames
-  uint32_t* frames = malloc(sizeof(uint32_t) * depth);
+  uint32_t *frames = malloc(sizeof(uint32_t) * depth);
   memset(frames, 0, sizeof(uint32_t) * depth);
   jerry_get_stacktrace_depth(frames, depth);
 
--- a/deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
+++ b/deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
@@ -148,26 +148,29 @@ ecma_new_standard_error (ecma_standard_error_t error_type) /**< native error typ
 
   uint32_t depth = 10;
   // create frames
-  uint32_t *frames = malloc(sizeof(uint32_t) * depth);
-  memset(frames, 0, sizeof(uint32_t) * depth);
-  jcontext_get_backtrace_depth(frames, depth);
+  size_t frames_mem_size = sizeof (uint32_t) * depth;
+  uint32_t *frames = malloc (frames_mem_size);
+  JERRY_ASSERT (frames != NULL);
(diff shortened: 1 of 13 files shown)


src/js/iotjs.js
```diff
@@ -54,7 +54,6 @@
     try {
       var loc = '/tmp/iotjs.' + process.pid;
       lines = fs.readFileSync(loc).toString().split('\n');
-      fs.unlinkSync(loc);
```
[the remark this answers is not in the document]
legendecas: If we do remove this dump file, we can only get file read at the first time and failed afterwards.
Since dump file would be written continuously while new JS files are read in, it shall not be cached yet, in this case.

src/js/iotjs.js
```diff
@@ -196,56 +197,41 @@
...
+      enumerable: false,
+      get: function () {
+        if (this.__stack__ === undefined) {
+          process._flushParserDumpFile();
+          this.__stack__ = makeStackTraceFromDump(this.__frames__);
+        }
+        return this.__stack__;
+      },
+    },
+    __stack__: {
```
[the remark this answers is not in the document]
legendecas: As `__stack__` is a backing field of `stack`, the naming might be ok.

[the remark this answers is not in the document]
yorkie: > As __stack__ is a backing field of stack

Suggestion: just using `_stack` for this purpose :)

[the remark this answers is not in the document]
yorkie: BTW, `__proto__` is not forwarding to `proto`, actually Object doesn't own a `proto` object.

src/js/iotjs.js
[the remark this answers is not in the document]
yorkie: This should be called when the program gets crashed, so dump file might not be used any more? Could you please point me out where we use this file in another place?

deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
```diff
@@ -134,6 +135,32 @@ ecma_new_standard_error (ecma_standard_error_t error_type) /**< native error typ
...
+  jerry_value_t jframes = jerry_create_array(depth);
+  for (uint32_t i = 0; i < depth; ++i) {
+    jerry_set_property_by_index(jframes, i, jerry_create_number(frames[i]));
+  }
+
+  free(frames);
+
+  prop_value_p->value = jframes;
+  ecma_deref_object (ecma_get_object_from_value (jframes));
+
```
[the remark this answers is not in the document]
legendecas: Since error stack trace has to be started with `JERRY_CONTEXT.vm_top_context_p`, a `jerry_*` api has to be called in this case.
Or is there any way can pass through `JERRY_CONTEXT` to access current top vm context?

src/js/iotjs.js
```diff
@@ -196,56 +197,41 @@
...
+      configurable: true,
+      writable: true,
+      enumerable: false,
+    },
+  };
+  Object.defineProperties(Error.prototype, stackPropertiesDescriptor);
```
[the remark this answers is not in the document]
legendecas: If the properties are not attached to the prototype, how can we get all error instance got the property `stack` without modifying native `Error` constructor and sharing the same stack generating code in JS?
The prototype changing of native types has to be prevented in cases of implementing our own API, however `error.stack` is a commonly used property in JS, it might be ok.

src/js/iotjs.js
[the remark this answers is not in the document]
legendecas: In this implementation, stacks are generated from parser dump file with an array of frames.
Like examples I stated in issue description, it is possible that multiple error instances' stack accessed in a single run. And since the scripts are loaded with `require` continuously, the parser dump file does get new contents too.
Thus it is required to not remove the dump file to get error stacks in newly loaded scripts be generated correctly.

deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
```diff
@@ -134,6 +135,32 @@ ecma_new_standard_error (ecma_standard_error_t error_type) /**< native error typ
...
+                                                                         frames_str_p,
+                                                                         ECMA_PROPERTY_CONFIGURABLE_WRITABLE,
+                                                                         NULL);
+  ecma_deref_ecma_string (frames_str_p);
+
+  uint32_t depth = 10;
+  // create frames
+  uint32_t *frames = malloc(sizeof(uint32_t) * depth);
+  memset(frames, 0, sizeof(uint32_t) * depth);
+  jerry_get_stacktrace_depth(frames, depth);
```
yorkie: `jerry_get_stacktrace_depth()` is defined at [here](https://github.com/Rokid/ShadowNode/blob/e6ced3744bf3bcc8d44ef90edaafebb545b5012c/deps/jerry/jerry-core/api/jerry.c#L3392), and I think we could move the implementation into `jcontext`, and in jerry-core/api/jerry.c, we just call the internal impl as the following:

```c
void
jerry_get_stacktrace_depth (uint32_t *frames, uint32_t depth)
{
  jerry_assert_api_available ();
  return JERRY_BACKTRACE_DEPTH(frames, depth);
}
```

deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
```diff
@@ -134,6 +135,32 @@ ecma_new_standard_error (ecma_standard_error_t error_type) /**< native error typ
...
+
+  jerry_value_t jframes = jerry_create_array(depth);
+  for (uint32_t i = 0; i < depth; ++i) {
+    jerry_set_property_by_index(jframes, i, jerry_create_number(frames[i]));
+  }
```
yorkie: We shouldn't call the high-level API, namely the `jerry_*` in its submodule like ecma, so may I ask you to change to use the relevant methods under `jerry-core/ecma`.

deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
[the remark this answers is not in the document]
yorkie: See the above comments of mine, I think in ecma, call jcontext looks acceptable. Then we need to move some code into [jcontext.h](https://github.com/Rokid/ShadowNode/blob/e6ced3744bf3bcc8d44ef90edaafebb545b5012c/deps/jerry/jerry-core/jcontext/jcontext.h), and call those from there :)

deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
yorkie: BTW, the stacktrace are not regular naming rule in JerryScript, let's use `backtrace` to keep consistent with the community :)

src/js/iotjs.js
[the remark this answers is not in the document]
yorkie: Ok, looks good to me now, and only one question we need to acquire, when do we clean this dump file up, in our past case, @lolBig give a feedback about this, when the program exits and restarts in a infinite loop, the dump file will be generated every circle, that eats the disk up to empty.

src/js/iotjs.js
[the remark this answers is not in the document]
yorkie: Good point, a false `enumerable` value is reasonable and you has added it in the latest patch :p

test/run_pass/test_iotjs_error_stack.js
```diff
@@ -0,0 +1,19 @@
+var assert = require('assert');
+
+var error = getError();
+assert(typeof error.stack === 'string');
+assert(error.stack.match(/^    at getError/));
```
yorkie: A little suggestion, could we changing this match rule to simple `/at GetError/`? That don't make me in trouble whenever we tweak the spaces :)

test/run_pass/test_iotjs_error_stack.js
```diff
@@ -0,0 +1,19 @@
...
+
+var plain = getPlainObject();
+assert(typeof plain.stack === 'string');
+assert(error.stack.match(/^    at getPlainObject/));
```
yorkie: Ditto :)

deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
```diff
@@ -134,6 +136,42 @@ ecma_new_standard_error (ecma_standard_error_t error_type) /**< native error typ
...
+
+  ecma_property_value_t *prop_value_p = ecma_create_named_data_property (new_error_obj_p,
+                                                                         frames_str_p,
+                                                                         ECMA_PROPERTY_CONFIGURABLE_WRITABLE,
+                                                                         NULL);
+  ecma_deref_ecma_string (frames_str_p);
+
+  uint32_t depth = 10;
+  // create frames
+  uint32_t *frames = malloc(sizeof(uint32_t) * depth);
```
yorkie: Should check the result of `malloc()` like the following:

```c
uint32_t *frames = malloc (frames_size);
if (frames == NULL)
{
  // throw errors
}
memset (frames, 0, frames_size);
```

deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
```diff
@@ -134,6 +136,42 @@ ecma_new_standard_error (ecma_standard_error_t error_type) /**< native error typ
...
+    ecma_string_t *str_idx_p = ecma_new_ecma_string_from_uint32 ((uint32_t) i);
+    ecma_property_value_t *index_prop_value_p = ecma_create_named_data_property (array_p,
+                                                                           str_idx_p,
+                                                                           ECMA_PROPERTY_CONFIGURABLE_WRITABLE,
+                                                                           NULL);
+    ecma_deref_ecma_string (str_idx_p);
+    index_prop_value_p->value = ecma_make_uint32_value(frames[i]);
+  }
+
+  free(frames);
```
yorkie: Use `JERRY_ASSERT` to make sure the frames are malloc-ed successfully :)

deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
```diff
@@ -134,6 +136,42 @@ ecma_new_standard_error (ecma_standard_error_t error_type) /**< native error typ
...
+  uint32_t *frames = malloc(sizeof(uint32_t) * depth);
+  memset(frames, 0, sizeof(uint32_t) * depth);
+  jcontext_get_backtrace_depth(frames, depth);
+
+  ecma_value_t frames_array_length_val = ecma_make_uint32_value (depth);
+  ecma_value_t frames_array = ecma_op_create_array_object (&frames_array_length_val, 1, true);
+  ecma_free_value (frames_array_length_val);
+  ecma_object_t *array_p = ecma_get_object_from_value (frames_array);
+
+  for (uint32_t i = 0; i < depth; ++i) {
```
yorkie: Let's follow the coding style of JerryScript, put the left-bracket `{` on the next new line.

deps/jerry/jerry-core/jcontext/jcontext.c
```diff
@@ -51,6 +51,18 @@ jerry_hash_table_t jerry_global_hash_table;
 #endif /* !CONFIG_ECMA_LCACHE_DISABLE */
 #endif /* !JERRY_ENABLE_EXTERNAL_CONTEXT */
 
+void
+jcontext_get_backtrace_depth (uint32_t *frames, uint32_t depth) {
```
yorkie: Please fix the coding style as well :)

deps/jerry/jerry-core/jcontext/jcontext.c
```diff
@@ -51,6 +51,18 @@ jerry_hash_table_t jerry_global_hash_table;
...
+  vm_frame_ctx_t *ctx_p = JERRY_CONTEXT (vm_top_context_p);
+  for (uint32_t idx = 0; idx < depth && ctx_p != NULL; ++idx) {
```
yorkie: Ditto.

src/js/iotjs.js
```diff
@@ -196,56 +194,41 @@
...
-      CallSite.create('main', ['<anonymous>', 4, 7]),
-    ];
+    return makeStackTraceFromDump(throwable._frames || []);
   }
 
+  var stackPropertiesDescriptor = {
+    stack: {
+      configurable: true,
+      enumerable: false,
+      get: function () {
```
yorkie: Tweak to `get: function() {` :)

src/js/iotjs.js
```diff
@@ -196,56 +194,41 @@
...
-    Object.defineProperties(throwable, {
-      stack: {
-        configurable: true,
-        get: function () {
-          return prepareStackTrace(throwable);
-        }
-      },
-      cachedStack: {
+    var frames = process._getStackFrames(11).slice(1);
+    Object.defineProperties(throwable, Object.assign({
```
yorkie: Could we split the complex expression like this:

```js
var descriptor = Object.assign({
  _frames: {
    configurable: true,
    writable: true,
    enumerable: false,
    value: frames,
  },
}, stackPropertiesDescriptor);
Object.defineProperties(throwable, descriptor);
```

That might be more clear for maintainer :)

yorkie: BTW, about the descriptor properties, 

- `configurable` might be false is OK?
- `writable` should be false? I didn't see any place we make re-assignment on `this._frames`.

yorkie: Why user need to change this value?

deps/jerry/jerry-core/ecma/operations/ecma-exceptions.c
```diff
@@ -134,6 +136,45 @@ ecma_new_standard_error (ecma_standard_error_t error_type) /**< native error typ
...
+  ecma_free_value (frames_array_length_val);
+  ecma_object_t *array_p = ecma_get_object_from_value (frames_array);
+
+  for (uint32_t i = 0; i < depth; ++i)
+  {
+    ecma_string_t *str_idx_p = ecma_new_ecma_string_from_uint32 ((uint32_t) i);
+    ecma_property_value_t *index_prop_value_p = ecma_create_named_data_property (array_p,
+                                                                                 str_idx_p,
+                                                                                 ECMA_PROPERTY_CONFIGURABLE_WRITABLE,
+                                                                                NULL);
```
yorkie: 2 spaces missed :(

legendecas:
Looks like CI failed for `uncaughtException handler throws: AssertionError: {"actual":"connect failed - status: ETIMEDOUT: developer.rokid.com","expected":"get did fail","operator":"==="}`. Is developer.rokid.com down?

yorkie:
Awesome works! Thanks @legendecas 💯

PR #123, opened 2018-05-18
jerry: record function name in bytecode

algebrait:


deps/jerry/jerry-core/ecma/base/ecma-globals.h
```diff
@@ -809,6 +809,7 @@ typedef struct
                                       *    the byte code is function or regular expression.
                                       *    If function, the other flags must be CBC_CODE_FLAGS...
                                       *    If regexp, the other flags must be RE_FLAG... */
+  ecma_value_t name;                /**< ecma value of function name */
```
yorkie: How about storing the name as string(`char*`) in clang?

deps/jerry/jerry-core/ecma/builtin-objects/ecma-builtin-function-prototype.c
```diff
@@ -69,6 +69,27 @@ ecma_builtin_function_prototype_object_to_string (ecma_value_t this_arg) /**< th
...
+      {
+        ecma_string_t *name_str_p = ecma_get_string_from_value (name);
+        lit_utf8_size_t sz = ecma_string_get_utf8_size (name_str_p);
+        lit_utf8_byte_t utf8_ptr[sz+1];
+        ecma_string_to_utf8_bytes (name_str_p, utf8_ptr, sz);
+        utf8_ptr[sz] = '\0';
+        printf("function name:%s\n", utf8_ptr);
+      }
+    }
+#endif
```
yorkie: Could we use test case instead?

deps/jerry/jerry-core/parser/js/js-parser-statm.c
```diff
@@ -472,7 +482,13 @@ parser_parse_function_statement (parser_context_t *context_p) /**< context */
 
   context_p->literal_count++;
 
-  lexer_construct_function_object (context_p, status_flags);
+  uint16_t index = lexer_construct_function_object (context_p, status_flags);
+
+  /* record function name in bytecode */
+  lexer_literal_t *func_literal = (lexer_literal_t *)parser_list_get(&context_p->literal_pool, index);
```
yorkie: Miss a space after `parser_list_get`.

deps/jerry/jerry-core/ecma/base/ecma-globals.h
yorkie: We also should declare how do we use this property, if we use it directly in JS land, ecma_value_t are proper than `char*`?

algebrait: Add internal property is also ok., if memory is acceptable.
ECMA spec does not say function should have name property, but implementer do. In V8,
`function foo(){] print(foo.name)`
will output foo.

yorkie: That's good feature, let's add it if memory is acceptable :)

yorkie: @algebrait could you add a test case for `foo.name`?

algebrait:
only pass in --no-snapshot mode now, do not merge

deps/jerry/jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -160,6 +160,18 @@ ecma_op_create_function_object (ecma_object_t *scope_p, /**< function's scope */
 
   /* 10., 11., 12. */
 
+  /* Add function name which is not in spec */
+  if(bytecode_data_p->name != ECMA_VALUE_EMPTY)
```
yorkie: Miss a space after the keyword `if` :)

yorkie:
> only pass in --no-snapshot mode now, do not merge

Seems fail as well :(

algebrait:
pass simple case of manually generated snapshot
`var a = function foo (){};
print(a.name);`
but still fail for iotjs

test/run_pass/test_function_name.js
```diff
@@ -0,0 +1,31 @@
...
+}
+
+assert.equal(foo.name, "foo");
+
+var var1 = function foo1() {
+}
+
+assert.equal(var1.name, "foo1");
+
+var var2 = ()=>{};
```
yorkie: Style fixes to: `var var2 = () => false;`

yorkie:
After running the function name enabled ShadowNode, it increases 100KB with 1M size codebase, let's merge it with this acceptable result :)

PR #128, opened 2018-05-22
build: add clang/cmake/shell/js linters and fixups

legendecas:
There are 2000+ issues of style and syntax currently.

Before starting working on fixing lint issues, the lint rules have to be determined first.

Which lint rule set shall we follow, [standard](standardjs.com), [airbnb style](https://github.com/airbnb/javascript), or [google style](https://google.github.io/styleguide/jsguide.html)?


yorkie:
Airbnb +1

/cc @Rokid/node-js

yorkie:
Let's just keep consistent with Node.js coding style.

PR #138, opened 2018-06-07
iotjs: call srand to set random seed

yorkie:
This fixes the `Math.random()` outputs the same order result in different process.

```js
for (var i = 0; i < 10; i++)
  console.log(Math.random());
```

Always output:

```
*.000007826369255781174
*.1315377880819142
*.755605321843177
*.4586501317098737
*.5327672371640801
*.21895918622612956
*.04704461619257927
*.6788647165521979
*.6792964055202902
*.9346928955055773
```

### DIFF
--- a/src/iotjs.c
+++ b/src/iotjs.c
@@ -28,6 +28,7 @@
 #include "jerryscript.h"
 
 #include <stdio.h>
+#include <stdlib.h>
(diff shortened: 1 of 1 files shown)


yorkie:
@lolBig @legendecas thanks for catching this bug :)

PR #140, opened 2018-06-08
console: optimize the log function for single param

yorkie:
BTW, added the benchmark for `console.log`.

```sh
> console/common.js n=128: 22,295.029427697045   #before
> console/common.js n=128: 38,630.9784834504     #after
```

The performance under this case is optimized 57% up :)

### DIFF
--- a/benchmark/console/common.js
+++ b/benchmark/console/common.js
@@ -1,0 +1,16 @@
+'use strict';
+
+var common = require('../common.js');
+var bench = common.createBenchmark(main, {
+  n: [128],
+});
+
+function main(opts) {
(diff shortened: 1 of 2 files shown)


legendecas:
There is no stack trace if an error instance is coerced to string, if it's more friendly that stack trace could be printed if an error is the only argument of invocations of console.log/info/... etc.?

yorkie:
Yea, you are write @legendecas, I'm glad that we could handle the single arguments as a special case in `util.format()`, could you please create a PR for this?

yorkie:
After optimizing the format function with single param, we could remove this check from here :)

PR #142, opened 2018-06-12
http: set content-length if not set and body is set

### LINKED ISSUE #102: http: set content-length by body

yorkie:
This fixes #102, now we have http/https is to set `Content-Length` by default if body is present. 

/cc @legendecas @lolBig


legendecas:
If `transfer-encoding` is `chunked`, which is default transfer encoding of node's http.request, header `content-length` may not be set implicitly, as it would be omitted by server.

yorkie:
@legendecas fixed and added test case :)

legendecas:
LGTM

PR #144, opened 2018-06-14
iotjs: check if uv_handle is valid for handlewrap safely

### LINKED ISSUE #143: iotjs: unexpected crash in core
The stack is:

```sh
#0  0x0000007faca34950 in uv_is_closing () from /usr/lib/libiotjs.so
[Current thread is 1 (Thread 0x7facef3010 (LWP 15097))]
(gdb) bt
#0  0x0000007faca34950 in uv_is_closing () from /usr/lib/libiotjs.so
#1  0x0000007fac9ed750 in iotjs_handlewrap_close () from /usr/lib/libiotjs.so
#2  0x0000007faca33f78 in uv_walk () from /usr/lib/libiotjs.so
#3  0x0000007fac9ec508 in iotjs_entry () from /usr/lib/libiotjs.so
#4  0x0000007fac71a8e8 in __libc_start_main (main=0x0, argc=0, argv=0x0,
    init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>,
    stack_end=<optimized out>) at libc-start.c:289
#5  0x0000000000400870 in _start () at ../sysdeps/aarch64/start.S:81
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

yorkie:
This should fixes #143

Just check the handle data's flag, if not a handlewarp, just
directly close the handle.

/cc @lolBig @algebrait

### DIFF
--- a/src/iotjs.c
+++ b/src/iotjs.c
@@ -179,10 +179,14 @@ static int iotjs_start(iotjs_environment_t* env) {
 
 
 static void iotjs_uv_walk_to_close_callback(uv_handle_t* handle, void* arg) {
-  iotjs_handlewrap_t* handle_wrap = iotjs_handlewrap_from_handle(handle);
-  IOTJS_ASSERT(handle_wrap != NULL);
-
-  iotjs_handlewrap_close(handle_wrap, NULL);
+  iotjs_handlewrap_t* handlewrap = iotjs_handlewrap_from_handle_unsafe(handle);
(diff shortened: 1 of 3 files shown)


algebrait:
LGTM.
Please describe the bug clearly in related issue or add a test case.

yorkie:
The `uv_walk` callback checks all unclosed handles, but the previous implementation treat all handles as the `handle_wrap`, this patch just checks correctly and close properly.

PR #146, opened 2018-06-15
stack: should print error.stack when not catching uncaughtException

yorkie:
We should use `error.stack` as the outputs when not catching `uncaughtException`, and add the message line in front of `.stack`.

### DIFF
--- a/src/js/iotjs.js
+++ b/src/js/iotjs.js
@@ -50,9 +50,12 @@
   function loadDumpIfExists() {
     var lines = [];
     try {
+      process._flushParserDumpFile();
       var loc = '/tmp/iotjs.' + process.pid;
       lines = fs.readFileSync(loc).toString().split('\n');
-    } catch (err) {}
+    } catch (err) {
+      console.error(`occurrs unkwnown error when loading dump: ${err.message}`);
+    }
     return lines;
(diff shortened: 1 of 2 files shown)


src/js/iotjs.js
```diff
@@ -203,8 +206,8 @@
       enumerable: false,
       get: function() {
         if (this.__stack__ === undefined) {
-          process._flushParserDumpFile();
-          this.__stack__ = makeStackTraceFromDump(this.__frames__ || []);
+          this.__stack__ = `Error: ${this.message}\n`
```
legendecas: The format of first line might be `${this.name}: ${this.message}\n` rather than a hardcoded `Error: xxx`.

legendecas: So an `TypeError` or any other `Error` inherited error class with it's own name can be printed as:
```
TypeError: unexpected string in arguments
   ...foobar stack trace
```

legendecas: Actually the `name` property of error instance is set by implementer manually rather than derived from the constructor's name.
yet LGTM🚀

yorkie: Maybe you forgot approve this Patch >.<

PR #147, opened 2018-06-15
iotjs: handle segv and exit event, remove dump file when exiting.

yorkie:
To test the case, added a kill method to monitor the segv/... signal.

### DIFF
--- a/src/iotjs.c
+++ b/src/iotjs.c
@@ -281,7 +281,6 @@ int iotjs_entry(int argc, char** argv) {
 
   // Release JerryScript engine.
   iotjs_jerry_release(env);
-  remove(JS_DUMP_PATHNAME);
 
 terminate:;
   bool context_reset = false;
--- a/src/iotjs.c
+++ b/src/iotjs.c
@@ -194,22 +194,22 @@ static void iotjs_onexit(void) {
   remove(JS_DUMP_PATHNAME);
 }
 
-static void iotjs_onsegv(int signum) {
+static void iotjs_onsignal(int signum) {
   remove(JS_DUMP_PATHNAME);
   signal(signum, SIG_DFL);
   kill(getpid(), signum);
 }
 
 static void atsignal() {
-  struct sigaction sa_segfault;
-  sa_segfault.sa_flags = 0;
-  sa_segfault.sa_handler = iotjs_onsegv;
-  sigaction(SIGQUIT, &sa_segfault, NULL);
-  sigaction(SIGILL, &sa_segfault, NULL);
-  sigaction(SIGABRT, &sa_segfault, NULL);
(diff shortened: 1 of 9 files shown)


yorkie:
To make the CI green, it depends on #146 to be merged, so suspend this review.

tools/check_crash.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env sh
+
+IOTJS_BIN=$1
+
+trap ':' SEGV
+
+for file in $(ls ./test/crash/*.js)
+do
+  echo "running $file"
+  $IOTJS_BIN $file > /dev/null 2>&1
```
legendecas: We can collect pid of each run
e.g.
```sh
pids="$(sh -c "echo \$\$; exec $IOTJS_BIN $file > /dev/null 2>&1") $pids"
```

then we can determine if the dump file of each process are removed rather than expecting no `/tmp/iotjs.*` exists

yorkie: Good point 👍

yorkie:
@legendecas fixed :)

src/iotjs.c
```diff
@@ -224,9 +247,13 @@ int iotjs_entry(int argc, char** argv) {
   iotjs_environment_set_loop(env, uv_default_loop());
 
   // set parser dump file
-  char dump_path[20];
-  snprintf(dump_path, 20, "/tmp/iotjs.%zu", (size_t)getpid());
```
algebrait: Macro is better than magic number, especially if appear more than once.

src/iotjs.c
```diff
@@ -189,6 +190,28 @@ static jerry_value_t dummy_wait_for_client_source_cb() {
...
+}
+
+static void iotjs_onsegv(int signum) {
+  remove(JS_DUMP_PATHNAME);
+  signal(signum, SIG_DFL);
+  kill(getpid(), signum);
+}
+
+static void atsignal() {
+  struct sigaction sa_segfault;
```
algebrait: name iotjs_onsegv and sa_segfault are confusing. In fact, you have installed many abnormal signal handler

src/iotjs.c
yorkie: What do you mean by the magic number?

src/iotjs.c
yorkie: Fixed, thanks :)

src/iotjs.c
algebrait: I mean 20. It's a good habit. If more than one magic number are far from each other, it's easy to forget sync all if want change value.
But your code is OK here.

yorkie: Yea, I have used `sizeof` instead of the numbers :)

PR #148, opened 2018-06-16
[WIP] Performance: implement a simpler CPU profiling API

yorkie:
This is based on the @algebrait's patch, and added some extra features then. Still working in progress, no reviewing for merge purpose for now :)


yorkie:
After taking a survey about profiling on Node.js ecosystem, I propose to add a builtin module `profiler` for `v8-profiler`, which is designed to support the following APIs:

- `takeSnapshot()`: for future heap stats.
- `startProfiling()`: start profiling for current process.
- `stopProfiling()`: stop profiling and get results information in some way.

yorkie:
This depends on #160 to be merged :)

Issue #79, opened 2018-03-15
https.setTimeout would throw an uncaughtException: TypeError: Expected a function.


lolBig:
rt

legendecas:
`tls.TLSSocket` do not inherit from `net.Socket` while in node.js does inherit from `net.Socket`. So there are batch of features not implemented in `tls.TLSSocket`.

yorkie:
@legendecas could you please change it consistent with node implementation? I did, but failed.

legendecas:
Resolved by #149

PR #149, opened 2018-06-20
tls: inherits TLS socket from net.Socket

legendecas:
Expect resolving #79


src/js/tls.js
```diff
@@ -3,85 +3,110 @@
...
-  if (!(this instanceof TLSSocket))
-    return new TLSSocket(socket, opts);
-  EventEmitter.call(this);
-
-  var tlsOptions = Object.assign({
-    servername: socket.host || socket.hostname,
-    // rejectUnauthorized: false,
-  }, opts);
+function delegates(proto, target) {
+  if (!(this instanceof delegates)) return new delegates(proto, target);
```
yorkie: New line for `return` statement please :)

src/js/tls.js
```diff
@@ -3,85 +3,110 @@
...
-function TLSSocket(socket, opts) {
-  if (!(this instanceof TLSSocket))
-    return new TLSSocket(socket, opts);
-  EventEmitter.call(this);
-
-  var tlsOptions = Object.assign({
-    servername: socket.host || socket.hostname,
-    // rejectUnauthorized: false,
-  }, opts);
+function delegates(proto, target) {
```
yorkie: Could we use `Delegate` as the class name?

src/js/tls.js
```diff
@@ -3,85 +3,110 @@
...
+  var proto = this.proto;
+  var target = this.target;
 
-  // handle the [ca1,ca2,...]
-  if (Array.isArray(tlsOptions.ca)) {
-    tlsOptions.ca = tlsOptions.ca.join('\n');
-  }
+  proto[name] = function delegation() {
+    return this[target][name].apply(this[target], arguments);
+  };
```
yorkie: Hmm, I don't like this way to delegate the methods from `net` to `tls`, which would name the function to `delegation`, could we just assign it?

Btw, why not `Object.assign` for the delegating functions?

src/modules/iotjs_module_tls.c
```diff
@@ -375,10 +378,21 @@ JS_FUNCTION(TlsEnd) {
   return jerry_create_undefined();
 }
 
+JS_FUNCTION(ErrName) {
+  DJS_CHECK_ARGS(1, number);
+
+  int errorcode = JS_GET_ARG(0, number);
+  char* err = malloc(sizeof(u_char) * 50);
```
yorkie: Could only alloc in stack? And miss free this block :)

src/js/tls.js
```diff
@@ -3,85 +3,110 @@
 require('crypto');
 
 var net = require('net');
+var tcp = require('tcp');
 var util = require('util');
-var EventEmitter = require('events').EventEmitter;
+/** @type tls.TlsWrap */
```
yorkie: What's the purpose of this line for?

yorkie:
Btw, I think this is a good start to talk about using TypeScript toolchain in ShadowNode, it looks good to me using typings firstly, and even writing builtin modules in TypeScript :)

@lolBig @algebrait How do you think about? Maybe implement a TypeScript interrupter at JerryScript layer?

algebrait:
interrupter -> interpreter?

I don't think implement typescript in jerryscript is a good way.
TS is a superset of the newest ES, which contains a complex parser and a heavy type infer system.
ES6 is already probably a little heavy for IoT device, not to mention TS.

But TS integration in shadow-node developing environment is an exciting feature. Type annotation will avoid many bugs, and help IDE auto-suggest and auto-complete. I even think almost all JS codes should be written in TS.

src/js/tls.js
legendecas: This annotation indicates what type of variable declared next line is

yorkie:
Okay, let's start from define the typings, but not this PR :)

typings/tls.d.ts
```diff
@@ -0,0 +1,23 @@
...
+  import { TLSSocket } from "tls";
+
+  interface TlsWrapOptions {
+    ca?: string
+    rejectUnauthorized?: boolean
+    servername?: string
+  }
+
+  class TlsWrap {
+    jsref?: TLSSocket
```
yorkie: Can we declare this in `private`?

src/modules/iotjs_module_tls.c
```diff
@@ -375,10 +378,21 @@ JS_FUNCTION(TlsEnd) {
   return jerry_create_undefined();
 }
 
+JS_FUNCTION(ErrName) {
+  DJS_CHECK_ARGS(1, number);
+
+  int errorcode = JS_GET_ARG(0, number);
+  char err[50];
```
yorkie: Actually we have the same function at https://github.com/Rokid/ShadowNode/blob/0c15a5f58cd8b551cef51aa4f1cd7273718e9e28/src/modules/iotjs_module_tls.c#L82, I think we need to make them work together :)

src/js/tls.js
```diff
@@ -3,85 +3,109 @@
...
+  });
+};
+
+TLSHandle.prototype.close = function close() {
+  this.tcpHandle.close();
+  this.tlsWrap.end();
 };
 
-TLSSocket.prototype.write = function(data, cb) {
+TLSHandle.prototype.readStart = function readStart() {};
```
yorkie: What is the function for? Some TODO comments should be added :)

src/js/tls.js
```diff
@@ -3,85 +3,109 @@
...
+  var proto = this.proto;
+  var target = this.target;
 
-  // Just a documented property to make secure sockets
-  // distinguishable from regular ones.
-  this.encrypted = false;
-  this._socket.on('error', this._tlsError.bind(this));
-  this._socket.on('connect', this.onsocket.bind(this));
-  this._socket.on('data', this.onsocketdata.bind(this));
+  proto[name] = this[target][name].bind(this[target]);
```
yorkie: Actually bond function still lost the function name :(

yorkie: How does just inherit from `net.Socket` like https://github.com/nodejs/node/blob/master/lib/_tls_wrap.js#L290-L341? I think this implementation is that we should follow yet :)

yorkie:
> But TS integration in shadow-node developing environment is an exciting feature. Type annotation will avoid many bugs, and help IDE auto-suggest and auto-complete. I even think almost all JS codes should be written in TS.

Yea, but we still have a challenge if the generated code from TypeScript could be ran well at our supported devices!

src/js/tls.js
legendecas: I'm afraid there is no way more elegant than this since it's required to access instance `this` in delegated method.
Do you have any suggestion to properly set up a proxy on prototype? 😋

src/js/tls.js
legendecas: `tls.TLSSocket` does inherit from `net.Socket` in this implementation, while `TLSHandle` doesn't. What `TLSHandle` do is to implement an interface that `net.Socket` interacts with.

src/modules/iotjs_module_tls.c
legendecas: The core of both function is `mbedtls_strerror(int, char*, size_t)`. Is it comprehensive to do another wrap over it?

src/js/tls.js
```diff
@@ -3,85 +3,114 @@
...
-  this._socket = new net.Socket(socket);
-  this._writev = [];
-  this._ended = false;
+function TLSHandle(options) {
+  var self = this;
+  this.tcpHandle = new tcp();
+  this.tlsWrap = new TlsWrap(options);
+  this.tlsWrap.onwrite = function onwrite(chunk) {
+    self.tcpHandle.write(chunk, function afterWrite(status) {
+    });
```
yorkie: Add a comment for empty function :)

src/js/tls.js
```diff
@@ -3,85 +3,114 @@
...
+    }
+  };
+  this.tcpHandle.onclose = this.tlsWrap.onclose = function onclose() {
+    if (self.onclose) {
+      self.onclose();
+    }
+  };
+
+  Object.defineProperty(this, 'owner', {
+    get: function get() {
```
yorkie: use `getOwner`.

src/js/tls.js
```diff
@@ -3,85 +3,114 @@
...
+      return self.tcpHandle.owner;
+    },
+    set: function set(val) {
```
yorkie: Ditto as `setOwner`.

src/js/tls.js
```diff
@@ -3,85 +3,114 @@
...
+    };
 
-  // bypass event emits
-  this._socket.on('close', onsocketclose.bind(this));
-  this._socket.on('finish', onsocketfinish.bind(this));
-  this._socket.on('end', onsocketend.bind(this));
+    self.tcpHandle.readStart();
+    var err = self.tlsWrap.handshake();
+    if (err < 0) {
+      throw TlsWrap.errname(err);
```
yorkie: Throwing a string is not recommended, maybe `new Error(TlsWrap.errname(code))`?

src/js/tls.js
legendecas: We have an eslint rule requires method name to be aligned with property name, i.e. we have to use `get` or alter our eslint rule.

PR #154, opened 2018-06-21
net: check if shutdown is a function.

### LINKED ISSUE #150: child_process: pipe has no shutdown function
The socket crashes at the line 526:

```js
// Writable stream finished.
function onSocketFinish() {
  var self = this;
  var state = self._socketState;

  if (!state.readable || self._readableState.ended || !self._handle) {
    // no readable stream or ended, destroy(close) socket.
    return self.destroy();
  } else {
    // Readable stream alive, shutdown only outgoing stream.
    self._handle.shutdown(function() {
      if (self._readableState.ended) {
        self.destroy();
      }
    });
  }
}
```

The reason is the handle, an `Pipe` object, doesn't own `shutdown` function.

yorkie:
Pipe doesn't own `shutdown` function, which causes crash when the writable is finished.

Fixes #150

### DIFF
--- a/src/js/net.js
+++ b/src/js/net.js
@@ -518,7 +518,9 @@ function onSocketFinish() {
   var self = this;
   var state = self._socketState;
 
-  if (!state.readable || self._readableState.ended || !self._handle) {
+  if (!state.readable || self._readableState.ended 
+    || !self._handle
+    || typeof self._handle.shutdown !== 'function') {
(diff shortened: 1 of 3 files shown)


yorkie:
Let's follow Node's implementation: https://github.com/nodejs/node/blob/master/lib/net.js#L356, @algebrait.

PR #155, opened 2018-06-21
iotjs: fix the compile failure when release build

yorkie:
In `Release` build doesn't has the `flag_create` option, we should find another way to validate if it is handwrap.


algebrait:
LGTM

PR #158, opened 2018-06-22
http: allow zero value for content-length by correct check for value

yorkie:
Allow the following:

```js
res.setHeader('content-length', 0);
```


legendecas:
Looks like CI failed for unable to connect developer.rokid.com 🌚

yorkie:
Yea, we need a local server instead of the remote one.

PR #157, opened 2018-06-22
build, travis: update travis build script

yorkie:
Add docker-based build and release types builds.

PR #164, opened 2018-06-26
 iotjs, env: add handlewrap_queue for clean it more clear

yorkie:
Fixed the build failure at #157 and a real fix to #143, it caused by the following error still:

```sh
/Users/travis/build/Rokid/ShadowNode/src/iotjs_handlewrap.c:136:20: error: comparison of address of 'handlewrap->unsafe' equal to a null pointer is always false [-Werror,-Wtautological-pointer-compare]
  if (&handlewrap->unsafe == NULL) {
       ~~~~~~~~~~~~^~~~~~    ~~~~
1 error generated.
```

### Why this patch?

This patch is related with how ShadowNode/IoT.js releases all the `HandleWrap` when program exits. The `HandleWrap` owns the `uv_handle_t` field. IoT.js releases them by `uv_walk` and its callback, which is called with an `uv_handle_t` object. 

The IoT.js asserts that every `uv_handle_t`'s `data` field is type of `HandleWrap`, That's working for these original modules, but now crashed for our new D-Bus module, that uses `uv_poll_t` and `uv_async_t` not as a `HandleWrap`.

I changed this to be checking if a `uv_handle_t` from a `HandleWrap` by its debug flag or `unsafe` field. After #157's failed build, We do think this implementation might be somehow not reliable! So the above is why this patch will be here.

### How to fix

In this patch, I'm doing:

- Remove the unneeded functions: `iotjs_handlewrap_validate_with_result()` and `iotjs_handlewrap_from_handle_unsafe()`.
- Introduce partially the library [clibs/list](https://github.com/clibs/list), a double linked list, as the `handlewrap` queue.
- In `handlewrap_initialize()`, push the instance into the above queue, and just remove it when destroy the handlewrap manually.
- Change the `uv_walk`'s callback only checks for `uv_handle_t`.
- Before `uv_walk`, clean up the queue of `HandleWrap`.

Therefore, we could clearly release these `HandleWrap`, and `uv_handle_t` as well. Request @algebrait @legendecas to have a review.

### DIFF
--- a/src/iotjs.c
+++ b/src/iotjs.c
@@ -180,14 +180,8 @@ static int iotjs_start(iotjs_environment_t* env) {
 }
 
 static void iotjs_uv_walk_to_close_callback(uv_handle_t* handle, void* arg) {
-  iotjs_handlewrap_t* handlewrap = iotjs_handlewrap_from_handle_unsafe(handle);
-  if (iotjs_handlewrap_validate_with_result(handlewrap)) {
-    iotjs_handlewrap_validate(handlewrap);
-    IOTJS_ASSERT(handlewrap != NULL);
-    iotjs_handlewrap_close(handlewrap, NULL);
-  } else if (uv_is_closing(handle) == 0) {
+  if (!uv_is_closing(handle))
     uv_close(handle, NULL);
-  }
 }
 
 static jerry_value_t dummy_wait_for_client_source_cb() {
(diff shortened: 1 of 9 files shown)


yorkie:
Oh, travis seems to be disabled by accident, just hold on.

yorkie:
Travis is back working :)

yorkie:
Ping @legendecas :)

Issue #133, opened 2018-05-30
http: support chunked transfer-encoding


legendecas:
By default, http module cannot not automatically derive content-length from input data since `ClientRequest` is a duplex stream, client can continuously write data into stream if the stream is not ended, yet headers have been sent at the beginning of the request. 
Thus the http client in node.js set header `Transfer-Encoding` with `chunked` by default, then encase data with an chunk length at the beginning of each chunk.

yorkie:
This has been fixed at #174.

PR #174, opened 2018-07-03
http: chunked transfer encoding

legendecas:
Expect resolving #133 

http client:
- [x] default encoding of `http` request shall be chunked
- [x] wrap http chunk with chunk header and a `\r\n` trailing

http server:
- [x] default encoding of response shall be chunked
- [x] wrap http chunk with chunk header and a `\r\n` trailing

------
- [x] all tests pass
- [x] test case over chunked encoding
- [x] document

### DIFF
--- a/src/js/http_client.js
+++ b/src/js/http_client.js
@@ -221,7 +221,7 @@ function responseOnEnd() {
 }
 
 ClientRequest.prototype._implicitHeader = function _implicitHeader() {
-  if (this._header) {
+  if (this._sentHeader) {
     throw new Error('Cannot render headers after they are sent');
   }
   this._storeHeader(this.method + ' ' + this.path + ' HTTP/1.1\r\n');
--- a/src/js/http_client.js
+++ b/src/js/http_client.js
@@ -29,7 +29,7 @@ function ClientRequest(options, cb) {
   var port = options.port = options.port || 80;
   var host = options.host = options.hostname || options.host || '127.0.0.1';
   var method = options.method || 'GET';
-  var path = options.path || '/';
+  this.path = options.path || '/';
 
   var methodIsString = (typeof method === 'string');
   if (methodIsString && method) {
@@ -64,9 +64,6 @@ function ClientRequest(options, cb) {
     }
     this.setHeader('Host', hostHeader);
   }
-
-  // store first header line to be sent.
-  this._storeHeader(method + ' ' + path + ' HTTP/1.1\r\n');
 
   // Register response event handler.
   if (cb) {
@@ -222,6 +219,13 @@ function responseOnEnd() {
     socket.destroySoon();
   }
 }
+
+ClientRequest.prototype._implicitHeader = function _implicitHeader() {
+  if (this._header) {
+    throw new Error('Cannot render headers after they are sent');
+  }
+  this._storeHeader(this.method + ' ' + this.path + ' HTTP/1.1\r\n');
+};
 
 ClientRequest.prototype.abort = function() {
   var self = this;
--- a/src/js/http_client.js
+++ b/src/js/http_client.js
@@ -31,6 +31,23 @@ function ClientRequest(options, cb) {
   var method = options.method || 'GET';
   var path = options.path || '/';
 
+  var methodIsString = (typeof method === 'string');
+  if (methodIsString && method) {
+    method = this.method = method.toUpperCase();
+  } else {
+    method = this.method = 'GET';
+  }
+
+  if (method === 'GET' ||
+      method === 'HEAD' ||
+      method === 'DELETE' ||
+      method === 'OPTIONS' ||
+      method === 'CONNECT') {
+    this.useChunkedEncodingByDefault = false;
+  } else {
+    this.useChunkedEncodingByDefault = true;
+  }
+
   // If `options` contains header information, save it.
   if (options.headers) {
     var keys = Object.keys(options.headers);
--- a/src/js/http_outgoing.js
+++ b/src/js/http_outgoing.js
@@ -26,7 +26,7 @@ function OutgoingMessage() {
   this.writable = true;
 
   this._hasBody = true;
-  this._contentLength = 0;
+  this._contentLength = null;
   this.chunkedEncoding = false;
   this.useChunkedEncodingByDefault = true;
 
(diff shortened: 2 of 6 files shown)


src/js/http_client.js
```diff
@@ -29,7 +29,24 @@ function ClientRequest(options, cb) {
   var port = options.port = options.port || 80;
   var host = options.host = options.hostname || options.host || '127.0.0.1';
   var method = options.method || 'GET';
-  var path = options.path || '/';
+  this.path = options.path || '/';
+
+  var methodIsString = (typeof method === 'string');
+  if (methodIsString && method) {
```
yorkie: I think we just defines all methods and check by `methods[method]` here.

src/js/http_client.js
```diff
@@ -29,7 +29,24 @@ function ClientRequest(options, cb) {
...
+
+  if (method === 'GET' ||
+      method === 'HEAD' ||
+      method === 'DELETE' ||
+      method === 'OPTIONS' ||
+      method === 'CONNECT') {
+    this.useChunkedEncodingByDefault = false;
+  } else {
+    this.useChunkedEncodingByDefault = true;
+  }
```
yorkie: I'm going to make this to be a stateless function like this:

```js
function isChunkedByDefault(req) {
  if (['GET', 'HEAD', ...].indexOf(req.method) !== -1) {
    return false;
  } else {
    return true;
  }
}
```

src/js/http_client.js
```diff
@@ -206,6 +220,13 @@ function responseOnEnd() {
   }
 }
 
+ClientRequest.prototype._implicitHeader = function _implicitHeader() {
+  if (this._header) {
+    throw new Error('Cannot render headers after they are sent');
```
yorkie: This code seems say: if header does not exist, it's unable to render again, because it can be sent? Why the `this._header` means the response has been sent?

src/js/http_outgoing.js
```diff
@@ -17,14 +17,18 @@
 var util = require('util');
 var stream = require('stream');
 
+var crlf_buf = Buffer.from('\r\n');
```
yorkie: Use `crlf` or `crlfBuf`?

src/js/http_outgoing.js
```diff
@@ -133,15 +141,33 @@ OutgoingMessage.prototype.write = function(chunk, encoding, callback) {
...
+  var len, ret;
+  if (this.chunkedEncoding) {
+    if (typeof chunk === 'string') {
+      len = Buffer.byteLength(chunk, encoding);
+    } else {
+      len = chunk.length;
+    }
 
-  return ret;
+    this._send(len.toString(16), 'latin1', null);
```
yorkie: We might have no "latin1" encoding yet.

src/js/http_outgoing.js
```diff
@@ -17,14 +17,18 @@
...
+
 
 function OutgoingMessage() {
   stream.Stream.call(this);
 
   this.writable = true;
 
   this._hasBody = true;
-  this._bodyLength = 0;
+  this._contentLength = null;
```
yorkie: Why removing the `_bodyLength`?

legendecas: It's renamed for align with naming of the header `Content-Length`.

yorkie: Good explain! 👍

src/js/http_client.js
legendecas: Actually it is possible to use a custom method rather than most common methods like GET, POST in http request.
#ref: https://tools.ietf.org/html/rfc2616#section-5.1.1

src/js/http_outgoing.js
legendecas: An alias of `latin1` is `binary`, which is one-byte encoded. 
I've tested over this, data have been correctly encoded.

src/js/http_client.js
yorkie: I think we don't need to support the whole context-free grammar for `Method`, a reduced CFG is:

```
Method         = "OPTIONS"                ; Section 9.2
                      | "GET"                          ; Section 9.3
                      | "HEAD"                       ; Section 9.4
                      | "POST"                       ; Section 9.5
                      | "PUT"                         ; Section 9.6
                      | "DELETE"                   ; Section 9.7
                      | "TRACE"                     ; Section 9.8
                      | "CONNECT"               ; Section 9.9
```

> An origin server SHOULD return the status code 405 (Method Not Allowed) if the method is known by the origin server but not allowed for the requested resource, and 501 (Not Implemented) if the method is unrecognized or not implemented by the origin server.

As the above description, we could respond a 501 for those methods is invalid.

src/js/http_outgoing.js
yorkie: Nope, see the following function: https://github.com/Rokid/ShadowNode/blob/3d408ea40191c693e470a9d0a48be60a98b84f38/src/js/net.js#L160-L165

It shows the encoding `latin1` could not be used.

legendecas: looks like it doesn't support specify encoding anyway. but for better compatibility in the future implementations, it's reasonable to leave as it is.

yorkie: The `write(chunk, callback)` only has 2 arguments, however we pass it with 3 arguments. Therefore this makes confusion and inconsistent, let's do that after we changed the function is 100% same as Node.js.

src/js/http_client.js
yorkie: Hmm, really think there are not too much developers to use the custom feature. But let's support it, that's reasonable than me.

src/js/http_outgoing.js
yorkie: Maybe we support encoding in `.write()` method before this?

legendecas: 😅Actually the chunk header is a hex number so that's ok to use default ascii encoding. 
It's ok to remove the encoding argument and use default ascii encoding at this scenario.

yorkie:
How is this in progress @legendecas? Currently I think this should be lack of only the documents, because of #184 did fix the encoding params.

legendecas:
Document is update! 🤠

yorkie:
Merged @legendecas :)

PR #176, opened 2018-07-03
profiling: add tool for CPU profiler

yorkie:
Work in progress...


algebrait:
If want dump snapshot debug info, try pseudo code below.
`  ecma_object_t *obj_iter_p = JERRY_CONTEXT (ecma_gc_objects_p);
  while (obj_iter_p != NULL)
  {
    dump bytecode compressed pointer if obj_iter_p is a function object and is not builtin
    obj_iter_p = ECMA_GET_POINTER (ecma_object_t, obj_iter_p->gc_next_cp);
  }`
But where to do it? Perhaps after snapshot data loaded is OK.

algebrait:
Parser dump file has been refactored, so jerry_stop_cpu_profiling should be re-implemented.

deps/jerry/jerry-core/api/jerry.c
```diff
@@ -3505,6 +3510,83 @@ jerry_get_backtrace_max_depth (void)
...
+    return false;
+  }
+
+  if (JERRY_CONTEXT (cpu_profiling_fp))
+  {
+    fclose (JERRY_CONTEXT (cpu_profiling_fp));
+    JERRY_CONTEXT (cpu_profiling_fp) = NULL;
+  }
+
+  jerry_flush_parser_dump_file ();
```
yorkie: This function seems to be removed from the latest tree.

yorkie: To read the dump file in clang, use https://github.com/Rokid/ShadowNode/blob/master/deps/jerry/jerry-core/api/jerry.c#L3447, and in JavaScript, see https://github.com/Rokid/ShadowNode/blob/master/src/js/iotjs.js#L53-L60.

src/js/profiler.js
```diff
@@ -0,0 +1,108 @@
...
+  } else if (arguments.length === 2) {
+    path = arguments[0];
+    duration = arguments[1];
+  }
+
+  native.startProfiling(path, duration);
+}
+
+function stopProfiling() {
+  native.stopProfiling();
```
yorkie: Suggestion: use `process._readParserDump()` here to make parser dump in persistent, such that can remove [these lines](https://github.com/Rokid/ShadowNode/blob/0e91917c138b1b5dfd5505f691524ee58190dddf/deps/jerry/jerry-core/api/jerry.c#L3581-L3587).

yorkie:
BTW, we need docs about how to use it at https://github.com/Rokid/ShadowNode/tree/master/docs/devs, too.

yorkie:
Looks good to me :)

PR #178, opened 2018-07-04
jerry, iotjs: refactor the parser dump to avoid the such many dump files

yorkie:
We still get some devices running [ShadowNode](https://github.com/Rokid/ShadowNode) which has huge parser dump files on temp directory after https://github.com/Rokid/ShadowNode/commit/4dcbb76e9f1eef13ed5dceac0ace7e7f940de485. The problem seems not handling the SIGKILL and others, there still are possibilities to leave these files.

See http://man7.org/linux/man-pages/man3/tmpfile.3.html:

> The tmpfile() function opens a unique temporary file in binary
> read/write (w+b) mode.  The file will be automatically deleted when
> it is closed or the program terminates.

This patch uses `tmpfile()` instead of `fopen` with a generated filename. And we do not need clean it any more.

Needs reviews from @algebrait @legendecas. Thank you :)


tools/check_crash.sh
```diff
@@ -9,13 +9,4 @@ do
   echo "running $file"
   pids="$(sh -c "echo \$\$; exec $IOTJS_BIN $file > /dev/null 2>&1") $pids"
```
legendecas: This test file seems meaningless if dump file removal check has been removed.

src/iotjs.c
```diff
@@ -33,7 +33,7 @@
 #include <sys/types.h>
 #include <unistd.h>
 
-static char JS_DUMP_PATHNAME[50];
+// static char JS_DUMP_PATHNAME[50];
```
yorkie: This needs to be removed, too.

yorkie:
Fixed @legendecas

test.js
```diff
@@ -0,0 +1,4 @@
+
+setInterval(() => {
```
legendecas: looks like this is a temp test file 😅

yorkie: Yea, removed aha.. :p

yorkie:
CI is green, let's merge this.

PR #184, opened 2018-07-06
stream: writable#write shall respect chunk encoding

legendecas:
writable#write shall pass chunk encoding to underlying stream implementation

- [x] document

## Effect on API Stability

3 api have changed in this PR. However the arguments order in 2 of them are not necessarily forced in order of `chunk, encoding, callback`, as second argument encoding is optional. 

A break change is introduced in this PR to `Writable#_write`. The function signature has changed from
```typescript
Writable#_write(chunk: Buffer | string, callback: Function, onwrite: Function)
```
to 
```typescript
Writable#_write(chunk: Buffer | string, encoding: string, callback: Function, onwrite: Function)
```

### DIFF
--- a/docs/api/Stream.md
+++ b/docs/api/Stream.md
@@ -367,8 +367,9 @@ var writable = new Writable({ highWaterMark:256 });
 ```
 
 
-### writable.end([chunk[, callback]])
+### writable.end([chunk][, encoding][, callback])
 * `chunk` {Buffer|string} Final data to write.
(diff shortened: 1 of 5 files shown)


yorkie:
@legendecas create another issue to track the `objectMode` and `decodeStrings` options, and skip them in this PR, and needs to update the docs for encoding.

yorkie:
Merged and thank you @legendecas, BTW this follows Node.js API such that it should not be flagged as a breaking change.

PR #187, opened 2018-07-09
DBus: fix memory leaks

lolBig:


yorkie:
@lolBig May I ask you to cleanup those unnecessary commits and lines from this patch, so that we could see what's the real things to be changed.

yorkie:
Oh, you seems forking from the branch v0.8.x? Note that the v0.8.x is a release branch, not for development, so don't fork from it any more, please rebase your patches from the current `master`, thank you :)

yorkie:
Closed because of the new duplicated one #188. And remember that keep 1 branch for the same topic to avoid the duplication as possible.

PR #193, opened 2018-07-10
iotjs: improve the cli options, more like nodejs

yorkie:
Now we are closer to Node.js style command line tool:

```sh
> iotjs -v
v0.8.5
> iotjs --version
v0.8.5
> iotjs --help
Usage: iotjs [options] [script | script.js] [arguments]

Options:
  -v, --version              print version
  -h, --help                 print help
  --loadstat                 print the load statistics
  --memstat                  print the memory statistics
  --show-opcodes             print the opcodes of running script
  --start-debug-server       activate the debugger server

Environment variables:
NODE_DISABLE_COLORS          set to 1 to disable colors in the REPL
NODE_PATH                    ':'-separated list of directories
                             prefixed to the module search path

Documentation can be found at https://github.com/Rokid/ShadowNode/tree/master/docs
```

### DIFF
--- a/src/iotjs_debuglog.h
+++ b/src/iotjs_debuglog.h
@@ -38,14 +38,15 @@ extern const iotjs_prefix_t iotjs_debug_prefix[4];
 /*
  * test with "nohup iotjs > /my.log 2>&1 < /dev/null"
  */
-#define DBG_COLOR_PRINT(prefix) do {                                \
-  if (isatty(0) == 1) {                                             \
-    fprintf(iotjs_log_stream,                                       \
-      "\033[%sm%s\033[0m ", prefix.color, prefix.text);             \
-  } else {                                                          \
-    fprintf(iotjs_log_stream, "[%s] ", prefix.text);                \
-  }                                                                 \
-} while (0)
+#define DBG_COLOR_PRINT(prefix)                                     \
+  do {                                                              \
+    if (isatty(0) == 1) {                                           \
+      fprintf(iotjs_log_stream, "\033[%sm%s\033[0m ", prefix.color, \
+              prefix.text);                                         \
+    } else {                                                        \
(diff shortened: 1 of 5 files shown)


src/iotjs_debuglog.h
```diff
@@ -32,7 +32,7 @@ extern const char* iotjs_debug_prefix[4];
 #define IOTJS_DLOG(lvl, ...)                                        \
   do {                                                              \
     if (0 <= lvl && lvl <= iotjs_debug_level && iotjs_log_stream) { \
-      fprintf(iotjs_log_stream, "[%s] ", iotjs_debug_prefix[lvl]);  \
+      fprintf(iotjs_log_stream, "%s ", iotjs_debug_prefix[lvl]);    \
```
legendecas: maybe brackets are still needed as some tty may not supporting color rendering

yorkie: How to detect if a tty supports colors?

yorkie: Adding the brackets seems not resolving this problem, either. Because `COLOR_PRINT` still could not be proceed correctly.

yorkie: Did google out the `tput` as well, but I'm not going to put this tool on our system, that might be useless except us, another example is here: [chalk/supports-color](https://github.com/chalk/supports-color/blob/master/index.js), I saw it checks `env.TERM` and windows support.

So how about the following algorithm?

- check `isatty()`, if no, we just not print color any more.
- check `env.TERM` that follows [chalk/supports-color](https://github.com/chalk/supports-color/blob/master/index.js).
- windows terminal support.

And in this PR, we should only have the step 1 done is enough?

legendecas: LGTM

yorkie:
@legendecas May I ask you to have a review again? Thank you :p

legendecas:
LGTM

PR #197, opened 2018-07-12
tty: implement the tty.isatty()

yorkie:
Implements the Node.js method: https://nodejs.org/dist/latest-v10.x/docs/api/tty.html#tty_tty_isatty_fd with a little bit diff, this throws an `Error` when `fd` is invalid by checking the `errno`.


yorkie:
The main APIs are not ready now, so let's keep no document.

PR #203, opened 2018-07-16
iotjs: fix memory leak

algebrait:
After fix this, valgrind on ubuntu pass all. So --no-check-valgrind can be removed.


yorkie:
How about enabling `valgrind check` from this one, just removing [this line here](https://github.com/Rokid/ShadowNode/blob/master/tools/travis_script.py#L30) :p

PR #207, opened 2018-07-17
iotjs: change iotjs_jval_as_string to use utf8 buffers

yorkie:
Upstream commit: https://github.com/Samsung/iotjs/commit/68e84c78c9cee2979831c842e57976ed9ca73ea8.

The detailed case is here: https://github.com/Samsung/iotjs/issues/1562.

/cc @lolBig


yorkie:
Merged, thanks for reviews :)

PR #214, opened 2018-07-17
dbus: should throw better message when caller is invalid

yorkie:
/cc @LanFly

### DIFF
--- a/src/js/dbus.js
+++ b/src/js/dbus.js
@@ -387,6 +387,10 @@ function ServiceInterface(dbus, name, service) {
 ServiceInterface.prototype.makeCall = function(member, data) {
   var self = this;
   var metadata = self._methods[member];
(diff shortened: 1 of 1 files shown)


src/js/dbus.js
```diff
@@ -387,6 +387,10 @@ function ServiceInterface(dbus, name, service) {
 ServiceInterface.prototype.makeCall = function(member, data) {
   var self = this;
   var metadata = self._methods[member];
+  if (!metadata || typeof metadata.handler !== 'function')
+    throw new TypeError(
+      `Invalid caller, ${member} are not defined at interface`);
```
legendecas: Since `metadata.handler` is check if is a function, maybe it's more reasonable to present a message that indicates `${member} is not callable`

yorkie: Fixed, thanks :)

PR #215, opened 2018-07-17
promise: remove native and use the pure javascript impl

### LINKED ISSUE #209: promise: segfault on jerry_run_all_enqueued_jobs()
The crashed stack is the below:

```
#0  0x0000007f77e63938 in ecma_get_object_type () from /usr/lib/libiotjs.so
#1  0x0000007f77e69930 in ecma_op_function_call () from /usr/lib/libiotjs.so
#2  0x0000007f77e6a35c in ecma_process_promise_reaction_job ()
   from /usr/lib/libiotjs.so
#3  0x0000007f77e6a6a4 in ecma_process_all_enqueued_jobs ()
   from /usr/lib/libiotjs.so
#4  0x0000007f77e48324 in iotjs_entry () from /usr/lib/libiotjs.so
#5  0x0000007f77b748e8 in __libc_start_main (main=0x0, argc=0, argv=0x0,
    init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>,
    stack_end=<optimized out>) at libc-start.c:289
#6  0x0000000000400860 in _start () at ../sysdeps/aarch64/start.S:81
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

yorkie:
This removes Promise from JerryScript, and added a pure JavaScript implementation. This tries to fix #209 as a side-effect.


deps/jerry/jerry-core/profiles/es2015-subset.profile
```diff
@@ -1 +1 @@
-# Currently an empty profile.
+CONFIG_DISABLE_ES2015_PROMISE_BUILTIN
```
legendecas: Missing EOF here :P

yorkie:
@algebrait May I ask you to have a review, too :)

deps/jerry/jerry-core/profiles/es2015-subset.profile
yorkie: Fixed, thanks :p

src/js/promise.js
```diff
@@ -0,0 +1,229 @@
...
+  this._state = 0;
+  this._handled = false;
+  this._value = undefined;
+  this._deferreds = [];
+
+  doResolve(fn, this);
+}
+
+function handle(self, deferred) {
+  while (self._state === 3) {
```
algebrait: Use upper case global var instead of magic number 0, 1, 2, 3, or add comment to these numbers?

yorkie:
> Are all code written by yourself? If almost all are from 3rd party, plz add original LICENSE and COPYRIGHT.

Good suggestion, I have added license and repository URI at https://github.com/Rokid/ShadowNode/pull/215/commits/1dd8d646109cd8e5e364a04ba8d15029e75a6829

PR #218, opened 2018-07-18
v0.8.7 Release Proposal

yorkie:
This patch mainly fixes the following problems:

- Use pure JavaScript `Promise` implementation to avoid #209.
- Fix the argument memory leak in Child Process.

And features:

- Add CPU Profiling (Needs compile).
- D-Bus: support multiple arguments for signal.
- Remove deprecated JavaScript test driver by Python TestRunner.

Community:

- [shadow-node/ffi](https://github.com/shadow-node/ffi) access native library with pure JavaScript by @legendecas.

Needs reviews from @algebrait @lolBig @legendecas.


yorkie:
Thanks all for the reviews 🌟

PR #221, opened 2018-07-19
events: optimize the performance of emit over 5x

yorkie:
Optimize the common case for emit arguments slicing, and the following shows the benchmark results:

```js
// before
$ iotjs benchmark/run.js events

events/emit.js
events/emit.js n=256: 343.92192036295285

// after
iotjs benchmark/run.js events

events/emit.js
events/emit.js n=256: 1,771.9394219594446
```

This optimizes 5x for `emit(name[, arg1, arg2])`.

/cc @lolBig @legendecas

### DIFF
--- a/src/js/events.js
+++ b/src/js/events.js
@@ -25,7 +25,7 @@ module.exports = EventEmitter;
 module.exports.EventEmitter = EventEmitter;
 
 
-EventEmitter.prototype.emit = function(type) {
+EventEmitter.prototype.emit = function(type, arg1, arg2) {
   if (!this._events) {
     this._events = {};
   }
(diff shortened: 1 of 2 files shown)


src/js/events.js
```diff
@@ -42,8 +42,13 @@ EventEmitter.prototype.emit = function(type) {
 
   var listeners = this._events[type];
   if (util.isArray(listeners)) {
-    listeners = listeners.slice();
-    var args = Array.prototype.slice.call(arguments, 1);
+    var args;
+    switch (arguments.length) {
+      case 0: args = []; break;
```
legendecas: Shall switch `arguments.length` starts on 1?

yorkie: Then how do we handle the case 0?

src/js/events.js
```diff
@@ -42,8 +42,13 @@ EventEmitter.prototype.emit = function(type) {
 
   var listeners = this._events[type];
   if (util.isArray(listeners)) {
-    listeners = listeners.slice();
```
yorkie: I dunno why this slice does exist, @legendecas @lolBig do you guess the reason?

src/js/events.js
yorkie: Never mind, that's my bug 👍

src/js/events.js
yorkie: Hmmm, the CI failure seems to be caused by this deleted line.

lolBig: seems this slice is unnecessary

lolBig: on the other hand, the value of the listeners is only `Array` or `undefined`, can we use `if (listeners)` instead of `if (util.isArray(listeners))`,

yorkie:
Closing this just because of this is fake opts :(

PR #225, opened 2018-07-25
process: add .title to set process name and get name

yorkie:
Add the `process.title` getter and setter.


yorkie:
The memory gets increased by 700KB after setting the `process.title` on MacOS, it's weird a bit.

yorkie:
Fixed @lolBig, help me review again :)

yorkie:
Thanks for your reviews @lolBig @legendecas :p

PR #226, opened 2018-07-26
dbus: fix memory leaks

lolBig:


yorkie:
Merged with a reword message, please note that memory may has many types, but the message should display how do we fix this memory leak, don't only write "fix memory leak".

/cc @Rokid/nodejs

PR #233, opened 2018-07-27
jerry: record buildtin accessor name's magic id

algebrait:


algebrait:
With this PR, following test will pass:

```shell
tools/build.py --builddir=build/tests/jerry_tests
deps/jerry/tools/runners/run-test-suite.sh deps/jerry/build/tests/jerry_tests/bin/jerry deps/jerry/tests/jerry-test-suite/es51-profile-list
```

PR #235, opened 2018-07-27
jerry: record buildtin accessor name's magic id (#233)

algebrait:


yorkie:
Closed this because this is unneeded, i will pull this commit from master to v0.9 :)

PR #237, opened 2018-07-29
build: update travis key

yorkie:
This updates the travis key, let's just see what's going on the tag `test-travis-key2`.


yorkie:
Even thought the travis CI shows it skips the deployments, however the assets has been added on the tag page: https://github.com/Rokid/ShadowNode/releases/tag/test-travis-key2.

PR #238, opened 2018-07-30
jerry: naïve implementation of handle scope

### LINKED ISSUE #229: jerry: implement handle_scope API
The API should be in `jerry-ext` as an extension instead of its auto-release.

legendecas:
A naïve implementation of handle scope, close #229 

- [x] implementation
- [x] tests
- [x] documents

## Memory Pool
To prevent trapping into system calls frequently, memory pools may be introduced into this PR.

- [x] HandleScope: Fixed pre-initialized jerry value id array, appended by a dynamically allocated handle chain list in heap
- [x] HandleScope: Pre-initialized fixed-size handle scope array, appended by a dynamically allocated scope chain list in heap

## Why it's still naïve

A jerry value has to be explicitly added to a scope rather than an automatic implicit procedure.

## Possible Pitfalls

1. a jerry value get from a api which isn't wrapped with any handle scope has to be explicitly added to a scope.
2. a jerry value get from a api which is wrapped with a handle scope shall not be added to a scope explicitly


deps/jerry/tests/unit-ext/test-ext-handle-scope-escape.c
```diff
@@ -0,0 +1,77 @@
...
+{
+  .free_cb = native_free_cb,
+};
+
+static jerry_value_t
+create_object (void)
+{
+  jerryx_handle_scope scope;
+  jerryx_open_handle_scope (&scope);
+  jerry_value_t obj = jerryx_handle_add (jerry_create_object ());
```
yorkie: Make some macros to simplify this function call?

legendecas: 🤯Sadly I didn't come up with any possible macro patterns that can be used to add a statement after a declaration. 
Do you have any idea of such kind of macro pattern?

yorkie:
Add jerryscript-ext test case in our travis script?

legendecas:
> Add jerryscript-ext test case in our travis script?

jerryscript tests already includes all jerryscript-ext tests actually.

yorkie:
Nope, see https://github.com/Rokid/ShadowNode/blob/master/tools/travis_script.py#L57-L60, we run the unit tests only now :)

legendecas:
> Nope, see https://github.com/Rokid/ShadowNode/blob/master/tools/travis_script.py#L57-L60, we run the unit tests only now :)

Actually test cases of jerryscript-ext are included in jerryscript's, which means if we run unit tests of jerryscript, tests of jerryscript-ext have also been ran. 😁

yorkie:
> Actually test cases of jerryscript-ext are included in jerryscript's, which means if we run unit tests of jerryscript, tests of jerryscript-ext have also been ran. 😁

Get it 👍

deps/jerry/docs/13.EXT-REFERENCE-HANDLE-SCOPE.md
```diff
@@ -0,0 +1,106 @@
...
+  jerry_value_t obj = jerry_create_object ();
+  return obj;
+} /* create_object */
+
+static void
+test_handle_scope_val (void)
+{
+  jerryx_handle_scope scope;
+  jerryx_open_handle_scope (&scope);
+  jerry_value_t obj = jerryx_handle_add (create_object ());
```
yorkie: jerry_create_object()

deps/jerry/jerry-ext/handle-scope/handle-scope.c
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * Opens a new handle scope and attach it to current global scope as a child scope.
+ *
+ * @return status code, jerryx_handle_scope_ok if success.
+ */
+jerryx_handle_scope_status
+jerryx_open_handle_scope(jerryx_handle_scope *result)
+{
+  jerryx_handle_scope scope = malloc(sizeof(jerryx_handle_scope_t));
```
yorkie:
```
jerryx_handle_scope scope = malloc (sizeof (jerryx_handle_scope_t));
```

yorkie: I just have a performance concern on these `malloc`, and it creates somehow memory fragment, too. How about creating a configurable `jerryx_handle_scope` pool, and use one of them directly.

deps/jerry/jerry-ext/handle-scope/handle-scope.c
```diff
@@ -0,0 +1,192 @@
...
+   * Release all handles related to given scope and its child scopes
+   */
+  jerryx_handle_scope a_scope = scope;
+  do
+  {
+    jerryx_handle_scope_release_handles(a_scope->handle_ptr);
+    jerryx_handle_scope child = a_scope->child;
+    free(a_scope);
+    a_scope = child;
+  } while(a_scope != NULL);
```
yorkie: Style nits need to be fixed.

deps/jerry/jerry-ext/handle-scope/handle-scope.c
```diff
@@ -0,0 +1,192 @@
...
+    scope->parent->child = NULL;
+  }
+
+  /**
+   * Release all handles related to given scope and its child scopes
+   */
+  jerryx_handle_scope a_scope = scope;
+  do
+  {
+    jerryx_handle_scope_release_handles(a_scope->handle_ptr);
```
yorkie: Ditto.

deps/jerry/jerry-ext/handle-scope/handle-scope.c
```diff
@@ -0,0 +1,192 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdlib.h>
+#include "jerryscript-ext/handle-scope.h"
+
+const jerryx_handle_scope_t jerryx_handle_scope_root;
```
lolBig: add static keyword?

deps/jerry/jerry-ext/handle-scope/handle-scope.c
lolBig: in my opinion, you should not use handle scope if you invoke a function great frequency, so don't worry about it

yorkie:
> A jerry value has to be explicitly added to a scope rather than an automatic implicit procedure.

I think we can address it at N-API :)

yorkie:
By the way, I think we need a way to lock those objects in an opened handle scope, that says an object in a handle scope should not be in the GC list.

yorkie:
Besides, could we keep the naming rule with v8? See https://v8docs.nodesource.com/node-0.8/d3/d95/classv8_1_1_handle_scope.html.

deps/jerry/jerry-ext/handle-scope/handle-scope.c
yorkie: Hmm, once the napi always uses handle scope, so that's unacceptable if this leaves fragments in memory.

deps/jerry/jerry-ext/handle-scope/handle-scope.c
yorkie: Agreed :)

deps/jerry/docs/13.EXT-REFERENCE-HANDLE-SCOPE.md
```diff
@@ -0,0 +1,106 @@
...
+**Summary**
+It is often necessary to make the lifespan of handles shorter than the lifespan of a native method. Even though the native code could only use the most recent handle, all of the associated objects would also be kept alive since they all share the same scope.
+
+To handle this case, JerryScript HandleScope extension provides the ability to establish a new 'scope' to which newly created handles will be associated. Once those handles are no longer required, the scope can be 'closed' and any handles associated with the scope are invalidated. The methods available to open/close scopes are `jerryx_open_handle_scope` and `jerryx_close_handle_scope`.
+
+JerryScript only supports a single nested hierarchy of scopes. There is only one active scope at any time, and all new handles will be associated with that scope while it is active. Scopes must be closed in the reverse order from which they are opened. In addition, all scopes created within a native method must be closed before returning from that method.
+
+**Example**
+
+[doctest]: # (test="compile")
```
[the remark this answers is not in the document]
legendecas: It's used to generate tests over code examples that resides in markdown documents to ensure there is no flaws in public docs.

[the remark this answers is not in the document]
yorkie: Who will test it?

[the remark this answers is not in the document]
legendecas: Take a look at this amazing test generator 💃https://github.com/Rokid/ShadowNode/blob/master/deps/jerry/tests/unit-doc/CMakeLists.txt#L19 
It's included in JerryScript's test sets same as jerry-ext.

deps/jerry/jerry-ext/handle-scope/handle-scope-allocator.c
```diff
@@ -0,0 +1,196 @@
...
+
+
+/**
+ * Get root handle scope.
+ */
+jerryx_handle_scope_t *
+jerryx_handle_scope_get_root (void)
+{
+  return &kJerryXHandleScopeRoot;
+}
```
yorkie: `jerryx_handle_scope_get_current` and `jerryx_handle_scope_get_root` could be static and inlined.

deps/jerry/jerry-ext/handle-scope/handle-scope-allocator.c
```diff
@@ -0,0 +1,196 @@
...
+
+/**
+ * Determines if given handle scope is located in pre-allocated list.
+ */
+bool
+jerryx_handle_scope_is_in_prelist (jerryx_handle_scope_t *scope)
+{
+  return (kJerryXHandleScopePool.prelist <= scope) &&
+    (scope <= (kJerryXHandleScopePool.prelist + JERRY_X_HANDLE_SCOPE_PRELIST_SCOPE_COUNT - 1));
+}
```
yorkie: Static and inline, too.

deps/jerry/jerry-ext/handle-scope/handle-scope-allocator.c
```diff
@@ -0,0 +1,196 @@
...
+    jerryx_handle_scope_t *scope = kJerryXHandleScopePool.prelist + kJerryXHandleScopePool.count;
+    scope->handle_count = 0;
+    scope->handle_ptr = NULL;
+
+    kJerryXHandleScopeCurrent = scope;
+    kJerryXHandleScopePool.count += 1;
+    return scope;
+  }
+
+  jerryx_handle_scope_dynamic_t *scope = malloc (sizeof(jerryx_handle_scope_dynamic_t));
```
yorkie: Check if the malloc is NULL.

deps/jerry/jerry-ext/handle-scope/handle-scope.c
```diff
@@ -0,0 +1,284 @@
...
+
+/**
+ * Add given jerry value to the scope.
+ */
+void
+jerryx_create_handle_in_scope (jerry_value_t jval, jerryx_handle_scope scope)
+{
+  size_t handle_count = scope->handle_count;
+  scope->handle_count += 1;
+  if (handle_count < JERRY_X_HANDLE_SCOPE_PRELIST_HANDLE_COUNT)
```
yorkie: `scope->handle_count - 1` might be meaningful :)

deps/jerry/jerry-ext/handle-scope/handle-scope.c
```diff
@@ -0,0 +1,284 @@
...
+  {
+    scope->handle_prelist[handle_count] = jval;
+    return;
+  }
+  jerryx_handle_t *handle = malloc (sizeof(jerryx_handle_t));
```
yorkie: Do check the malloc is NULL, either.

deps/jerry/jerry-ext/include/jerryscript-ext/handle-scope.h
```diff
@@ -0,0 +1,88 @@
...
+  jerry_value_t handle_prelist[JERRY_X_HANDLE_SCOPE_PRELIST_HANDLE_COUNT];
+  size_t handle_count;
+  jerryx_handle_t *handle_ptr;
+};
+
+
+typedef struct jerryx_handle_scope_dynamic_t jerryx_handle_scope_dynamic_t;
+struct jerryx_handle_scope_dynamic_t {
+  jerry_value_t handle_prelist[JERRY_X_HANDLE_SCOPE_PRELIST_HANDLE_COUNT];
+  size_t handle_count;
```
yorkie: To improve the code readability, see https://github.com/Rokid/ShadowNode/blob/master/deps/libtuv/include/uv.h#L492, and use macros to simplify here.

deps/jerry/jerry-ext/include/jerryscript-ext/handle-scope.h
```diff
@@ -0,0 +1,88 @@
...
+  jerryx_handle_t *handle_ptr;
+  jerryx_handle_scope_dynamic_t *child;
+  jerryx_handle_scope_dynamic_t *parent;
+};
+
+typedef enum {
+  jerryx_handle_scope_ok = 0,
+
```
yorkie: Remove this empty line, please.

deps/jerry/jerry-ext/handle-scope/handle-scope.c
```diff
@@ -0,0 +1,284 @@
...
+    {
+      /**
+       * Escape handle to parent scope
+       * handle count has been decremented in `jerryx_hand_scope_escape_handle_from_prelist`
+       */
+      *result =jerryx_hand_scope_escape_handle_from_prelist (scope, found_idx);
+      return jerryx_handle_scope_ok;
+    }
+  }
+  while (0);
```
yorkie: Why do you wrap the block with a `while(0)`, is it for scoping?

deps/jerry/jerry-ext/handle-scope/handle-scope.c
```diff
@@ -0,0 +1,284 @@
...
+
+  if (scope->handle_count <= JERRY_X_HANDLE_SCOPE_PRELIST_HANDLE_COUNT)
```
yorkie: Could you please move this line up to the head of the function? The `scope` seems not to be changed before :)

deps/jerry/jerry-ext/handle-scope/handle-scope.c
```diff
@@ -0,0 +1,284 @@
...
+    return jerryx_handle_scope_mismatch;
+  }
+
+  bool found = false;
+  do
+  {
+    size_t found_idx = 0;
+    for (size_t idx = 0; idx < JERRY_X_HANDLE_SCOPE_PRELIST_HANDLE_COUNT; idx++)
+    {
+      if (escapee == scope->handle_prelist[idx])
```
yorkie: How about checking this in reverse order? The `handle_prelist` is a LIFO(stack), searching from tail might be always faster than head, right?

legendecas: There is no actually out order of handle list, since we don't have a mechanism for remove a handle from a scope. The escaping doesn't guarantee if the escapee is the last one that added to the scope either.
Or is there any potential that an escapee is more possibly added to the scope lastly?

deps/jerry/jerry-ext/handle-scope/handle-scope.c
legendecas: Right, it's for scoping.

deps/jerry/jerry-ext/include/jerryscript-ext/handle-scope.h
legendecas: It's a separator between healthy status and error status.

deps/jerry/jerry-ext/handle-scope/handle-scope.c
legendecas: This line cannot be moved upwards since it ensures the chain list is not empty. If the list is empty, there is no need to continue the procedure below.

deps/jerry/jerry-ext/handle-scope/handle-scope.c
legendecas: If `malloc` does return a NULL pointer, what error shall be thrown here?

deps/jerry/jerry-ext/handle-scope/handle-scope-allocator.c
```diff
@@ -0,0 +1,196 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdlib.h>
+#include "handle-scope-internal.h"
+
+static jerryx_handle_scope_t kJerryXHandleScopeRoot = {
```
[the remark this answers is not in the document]
legendecas: Are there any existing naming rule for static variables? I used a rule in ObjC naming style.

yorkie:
Merged, thanks for this patch @legendecas :)

PR #239, opened 2018-07-30
CMake: fix params

lolBig:


cmake/jerry.cmake
```diff
@@ -24,7 +24,6 @@ ExternalProject_Add(hostjerry
   INSTALL_COMMAND ""
   CMAKE_ARGS
     -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
-    -DCMAKE_C_COMPILER=/usr/bin/gcc
```
yorkie: Why removing this line? It's for the host build.

cmake/config/arm-linux.cmake
```diff
@@ -15,5 +15,6 @@
 set(CMAKE_SYSTEM_NAME Linux)
 set(CMAKE_SYSTEM_PROCESSOR armv7l)
 
-set(CMAKE_C_COMPILER arm-linux-gnueabihf-gcc)
+set(CMAKE_SYSROOT ${CMAKE_EXTERNAL_SYSROOT})
+set(CMAKE_C_COMPILER ${CMAKE_TOOLCHAIN_ROOT}/arm-linux-gnueabihf-gcc)
```
yorkie: Actually if you wanna compile for Kamino, we should use https://github.com/Rokid/ShadowNode/blob/eec21ead1cfdc0857ff1a75cbc388a330c3ff058/cmake/config/openwrt-linux.cmake :)

yorkie:
@lolBig Could you please make the commit author are authorized by GitHub such that they could know you?

lolBig:
OK

PR #243, opened 2018-08-02
napi: module registration

### LINKED ISSUE #230: iotjs, addon: keep consistent with Node.js add-on loading
The current add-on loading are not consistent with Node.js, that might be required by N-API's module register.

legendecas:
close #230, depends on #247

NAPI module are loaded with gcc compilation attribute `__attribute__((constructor))`. Thus on the load process of a native add-on, a module registered with dylib constructor shall be loaded before current iotjs add-on module.

- [x] imported NAPI headers version management
- [x] tests
  - [x] ~~tests over if exports were been gc-ed~~ // jerry_value_t returned from `OpenNativeModule` is already acquired
  - [x] integrated test cli: run N-API testsets if --napi is enabled
- [x] ~~*documents*~~ (may be not required)
- [x] build NAPI by optional cli --napi option


include/node_api.h
```diff
@@ -0,0 +1,657 @@
+#ifndef SRC_NODE_API_H_
```
yorkie: How about downloading this from Node.js?

legendecas: Nay, IMHO it shall be included in our VSC to prevent unintended changes.

yorkie: I just think to sync these headers are boring work to anyone, to fix the unintended issue, we could lock the tag or commit id, only maintaining this a ID looks really simple than source code :)

yorkie:
We have LICENSE file in our project, so inline license head is unnecessary :)

legendecas:
> We have LICENSE file in our project, so inline license head is unnecessary :)

Most of source code in the project claimed with a license copy. Still, we'd better follow with current license convention.

tools/build.py
```diff
@@ -448,6 +467,7 @@ def run_checktest(options):
 
     # Run tests.
     if options.run_test:
+        build_napi_test_module(options)
```
yorkie: This should not be after the `run_test`.

legendecas: What if a compilation error occurred in napi test module? Shall it block main procedure? 🤔

yorkie:
> Most of source code in the project claimed with a license copy. Still, we'd better follow with current license convention.

Those are Samsung's license, we will remove it later.

tools/build.py
yorkie: https://github.com/Rokid/ShadowNode/pull/243/files/00c9ee210b3837abadf8ca30c65d2b691e2cd807#diff-d504d105fb41328f6722f4e5c8432a96L421 Put this line here?

legendecas:
@yorkie Just updated a simple script to pull node api headers from repository nodejs/node and prepend git hash to the pulled files.

src/modules/iotjs_module_process.c
```diff
@@ -413,7 +414,7 @@ JS_FUNCTION(ForceGC) {
   return jerry_create_boolean(true);
 }
 
-JS_FUNCTION(DLOpen) {
+JS_FUNCTION(OpenNativeModule) {
   iotjs_string_t location = JS_GET_ARG(0, string);
```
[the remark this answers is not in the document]
yorkie: Good catch, wow~

yorkie:
Merged and thanks for the awesome contribution!

PR #245, opened 2018-08-03
jerryx: root handle scope convenience methods

legendecas:
Series convenience methods of root handle scope.

Closes root handle scope on iotjs exit.

- [x] tests
- [x] add jerry-ext dependency to iotjs (#247 )
- [x] build option to include jerryx (#247 )

### DIFF
--- a/deps/jerry/jerry-ext/handle-scope/handle-scope-allocator.c
+++ b/deps/jerry/jerry-ext/handle-scope/handle-scope-allocator.c
@@ -75,6 +75,10 @@ jerryx_handle_scope_is_in_prelist (jerryx_handle_scope_t *scope)
 jerryx_handle_scope_t *
 jerryx_handle_scope_get_parent (jerryx_handle_scope_t *scope)
 {
+  if (scope == &kJerryXHandleScopeRoot)
+  {
+    return NULL;
+  }
   if (!jerryx_handle_scope_is_in_prelist (scope))
   {
     jerryx_handle_scope_dynamic_t *dy_scope = (jerryx_handle_scope_dynamic_t *) scope;
@@ -87,7 +91,7 @@ jerryx_handle_scope_get_parent (jerryx_handle_scope_t *scope)
   }
   if (scope == kJerryXHandleScopePool.prelist)
   {
-    return NULL;
+    return &kJerryXHandleScopeRoot;
   }
   return kJerryXHandleScopePool.prelist + JerryXHandleScopePrelistIdx (scope) - 1;
 }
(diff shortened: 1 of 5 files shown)


cmake/jerry.cmake
```diff
@@ -32,7 +32,7 @@ ExternalProject_Add(hostjerry
     -DJERRY_LIBC=OFF
     -DJERRY_CMDLINE=OFF
     -DJERRY_CMDLINE_SNAPSHOT=ON
-    -DJERRY_EXT=OFF
+    -DJERRY_EXT=ON
```
yorkie: Let's open it once the full functions are complete :)

cmake/jerry.cmake
```diff
@@ -52,7 +52,7 @@ macro(add_cmake_arg TARGET_ARG KEY)
 endmacro(add_cmake_arg)
 
 # Target libjerry
-set(JERRY_LIBS jerry-core jerry-port-default)
+set(JERRY_LIBS jerry-core jerry-port-default jerry-ext)
```
yorkie: We should set this by checking the define `JERRY_EXT`.

cmake/jerry.cmake
```diff
@@ -138,6 +138,7 @@ ExternalProject_Add(libjerry
 set_property(DIRECTORY APPEND PROPERTY
   ADDITIONAL_MAKE_CLEAN_FILES
     ${CMAKE_BINARY_DIR}/lib/libjerry-core.a
+    ${CMAKE_BINARY_DIR}/lib/libjerry-ext.a
```
yorkie: Ditto

src/iotjs.c
```diff
@@ -26,6 +26,7 @@
 #endif
 #include "jerryscript-port.h"
 #include "jerryscript.h"
+#include "jerryscript-ext/handle-scope.h"
```
yorkie: Also, we are not starting using HandleScope yet, so adding this by MACRO would be reasonable :)

src/iotjs.c
```diff
@@ -255,6 +256,7 @@ int iotjs_entry(int argc, char** argv) {
   }
 
   // Release JerryScript engine.
+  jerryx_close_handle_scope(jerryx_handle_scope_get_root());
```
yorkie: The same to this line :)

src/iotjs.c
legendecas: Once NAPI module are introduced to ShadowNode, handle-scope are necessary rather than configurable.

PR #246, opened 2018-08-03
module: any unknown extensions shall be treated as js

### LINKED ISSUE #244: module: unhandled condition of unknown extension
Unhandled condition of requiring files with unknown extension.

e.g.
```
$ iotjs foobar.c
// do nothing, with exit code 0
```

Shall files with unknown extension be loaded as if it is js files?

legendecas:
close #244 

- [x] test


yorkie:
@legendecas Add a test case under `/test/run_pass/require1`?

PR #247, opened 2018-08-03
cmake: add jerryx dependency to iotjs

legendecas:
depends on #245 

- [x] test option over jerryx enabled

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,8 +54,8 @@ jobs:
     - RUN_DOCKER=yes
   - stage: test
     env:
-    - JOBNAME="Linux/x86-64 without snapshot Build & Correctness Tests"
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -54,8 +54,8 @@ jobs:
     - RUN_DOCKER=yes
   - stage: test
     env:
-    - JOBNAME="Linux/x86-64 without snapshot Build & Correctness Tests"
-    - OPTS="no-snapshot"
```
yorkie: No-snapshot is required, too.

legendecas: There is a duplicated one of `no-snapshot`. Diff shows the duplicated one is replaced by the new jerryx job.

PR #249, opened 2018-08-03
console: improve the performance on printf

yorkie:
The original implementation would make a huge number of `syscalls` to `write()`.

/cc @legendecas

### DIFF
--- a/src/modules/iotjs_module_console.c
+++ b/src/modules/iotjs_module_console.c
@@ -20,19 +20,11 @@
 static jerry_value_t Print(const jerry_value_t* jargv,
                            const jerry_length_t jargc, FILE* out_fd) {
   JS_CHECK_ARGS(1, string);
-  iotjs_string_t msg = JS_GET_ARG(0, string);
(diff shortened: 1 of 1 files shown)


src/modules/iotjs_module_console.c
```diff
@@ -20,19 +20,11 @@
...
-    } else {
-      fprintf(out_fd, "\\u0000");
-    }
-  }
-  iotjs_string_destroy(&msg);
+  jerry_size_t len = jerry_get_string_size(jargv[0]);
+  jerry_char_t str[len + 1];
+  jerry_string_to_char_buffer(jargv[0], str, len);
+  str[len] = '\0';
+  fprintf(out_fd, "%s", str);
```
legendecas: Why we have to format one more time? If `fprintf(out_fd, str);` works?

legendecas: How can we reproduce the log which crashes the program?

yorkie: It's not the crash case, it causes the compile failing, do you have any patch here?

PR #256, opened 2018-08-07
handle-scope: complete escaping of handles

legendecas:
##### Checklist

- [x] `npm test` passes

There may be some handles we have to hand over it's life time management to jerry engine, and totally give up it's reference. Rather escaping it and keep a ref to the value in parent scope, the reference shall be deleted from the scope.

### DIFF
--- a/deps/jerry/jerry-ext/handle-scope/handle-scope-internal.h
+++ b/deps/jerry/jerry-ext/handle-scope/handle-scope-internal.h
@@ -76,6 +76,12 @@ jerryx_handle_scope_add_handle_to (jerryx_handle_t *handle, jerryx_handle_scope 
 
 void
 jerryx_create_handle_in_scope (jerry_value_t jval, jerryx_handle_scope scope);
+
+jerryx_handle_scope_status
+jerryx_escape_handle_impl (jerryx_escapable_handle_scope scope,
+                           jerry_value_t escapee,
+                           jerry_value_t *result,
+                           bool should_promote);
 /** MARK: - END handle-scope.c */
 
 #ifdef __cplusplus
--- a/deps/jerry/jerry-ext/handle-scope/handle-scope-internal.h
+++ b/deps/jerry/jerry-ext/handle-scope/handle-scope-internal.h
@@ -78,7 +78,7 @@ void
 jerryx_create_handle_in_scope (jerry_value_t jval, jerryx_handle_scope scope);
 
 jerryx_handle_scope_status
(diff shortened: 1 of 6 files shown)


deps/jerry/jerry-ext/handle-scope/handle-scope.c
```diff
@@ -231,18 +227,55 @@ jerryx_escape_handle (jerryx_escapable_handle_scope scope,
...
+
+
+/**
+ * Escape a handle from scope yet do not promote it to the outer scope.
+ * Leave the handle's life time management up to user.
+ *
+ * @return status code, jerryx_handle_scope_ok if success.
+ */
+jerryx_handle_scope_status
+jerryx_evade_handle (jerryx_escapable_handle_scope scope,
```
yorkie: What's the between `evade` with `escape`?

deps/jerry/jerry-ext/handle-scope/handle-scope-internal.h
```diff
@@ -76,6 +76,12 @@ jerryx_handle_scope_add_handle_to (jerryx_handle_t *handle, jerryx_handle_scope
 
 void
 jerryx_create_handle_in_scope (jerry_value_t jval, jerryx_handle_scope scope);
+
+jerryx_handle_scope_status
+jerryx_escape_handle_impl (jerryx_escapable_handle_scope scope,
```
yorkie: `jerryx_escape_handle_internal` might be proper than `impl`?

deps/jerry/jerry-ext/handle-scope/handle-scope-internal.h
```diff
@@ -76,6 +76,12 @@ jerryx_handle_scope_add_handle_to (jerryx_handle_t *handle, jerryx_handle_scope
 
 void
 jerryx_create_handle_in_scope (jerry_value_t jval, jerryx_handle_scope scope);
+
+jerryx_handle_scope_status
+jerryx_escape_handle_internal (jerryx_escapable_handle_scope scope,
+                           jerry_value_t escapee,
+                           jerry_value_t *result,
+                           bool should_promote);
```
yorkie: Alignments are required :(

legendecas: We have to turn of lint on jerry files :P

yorkie: Yea, just created an issue for this https://github.com/Rokid/ShadowNode/issues/258 :)

PR #272, opened 2018-08-09
module: optimize the compile, direct to read the source

yorkie:
##### Checklist

- [x] `npm test` passes

This directs to read the source inside the compile function, which in anyway decreases the objects in JerryScript, even thought the benchmark are not obvious.

### DIFF
--- a/src/modules/iotjs_module_process.c
+++ b/src/modules/iotjs_module_process.c
@@ -78,24 +78,30 @@ static jerry_value_t WrapEval(const char* name, size_t name_len,
 
 
 JS_FUNCTION(Compile) {
-  DJS_CHECK_ARGS(2, string, string);
+  DJS_CHECK_ARGS(1, string);
 
-  iotjs_string_t file = JS_GET_ARG(0, string);
-  iotjs_string_t source = JS_GET_ARG(1, string);
-
-  const char* filename = iotjs_string_data(&file);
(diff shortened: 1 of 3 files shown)


src/modules/iotjs_module_process.c
```diff
@@ -78,24 +78,33 @@ static jerry_value_t WrapEval(const char* name, size_t name_len,
...
+  uv_fs_stat(iotjs_environment_loop(env), &fs_req, filename, NULL);
+  uv_fs_req_cleanup(&fs_req);
+
+  if (!S_ISREG(fs_req.statbuf.st_mode)) {
+    iotjs_string_destroy(&path);
+    return JS_CREATE_ERROR(COMMON, "ReadSource error, not a regular file");
+  }
+
+  size_t size = 0;
+  char* source = iotjs__file_read(iotjs_string_data(&path), &size);
```
legendecas: Here and following previously extracted `filename` could be used for convenience and destroy `path` right after LINE 89. :p

src/modules/iotjs_module_process.c
```diff
@@ -119,9 +128,11 @@ JS_FUNCTION(CompileSnapshot) {
   size_t size = 0;
   char* bytecode = iotjs__file_read(iotjs_string_data(&path), &size);
```
legendecas: ditto.

yorkie:
Fixed @legendecas @algebrait :)

Issue #262, opened 2018-08-08
module: global module should be high priority than others


yorkie:
* **Version**: v0.9.0
* **Platform**: all
* **Subsystem**: module

Currently, the module searching path are local module is higher than global, that's incorrect. For instance, take a look at the following:

```
/usr/local/lib/node_modules/foobar/index.js
/opt/my_projects/foobar/index.js
/opt/my_projects/test.js
```

In `/opt/my_projects/test.js`, `require('foobar')` loads the `/opt/my_projects/foobar/index.js`.

legendecas:
Is this a path pattern issue? i.e. local packages must be loaded with relative path, `require('./foobar')`

yorkie:
Yea, IoT.js support this without the relative path dot, for compatibility, we have to tweak the priority at the first step :)

legendecas:
resolved with #274

PR #274, opened 2018-08-09
module: optimize the path searching

yorkie:
##### Checklist

- [x] `npm test` passes
- [x] tests and/or benchmarks are included(modified)

Give the following optimizations:

- search rules:
  - If the `id` is a relative path, directly search by `__dirname`, no global module and `node_modules`.
  - If the `id` is a global path, directly try with file.
- tweak the order of global paths, level up the `NODE_PATH`.
- remove the retry with `index` after package main property is absence, because checked already before.

Also, this patch removes the way to load relative module without dots:

```js
require('test/foobar') // doesn't work
require('./test/foobar') // it's working
```

### DIFF
--- a/docs/api/Module.md
+++ b/docs/api/Module.md
@@ -29,10 +29,10 @@ If a native module named `id` exists, load it and return.
 
 `require` function searches for modules in the following order:
 
-1. Current working directory.
-2. `iotjs_modules` folder under current working directory.
-3. `$HOME/iotjs_modules`
-4. `$IOTJS_PATH/iotjs_modules`
+1. `$NODE_PRIORITIZED_PATH/node_modules` if present.
+1. `node_modules` folder under current working directory.
+2. `$NODE_PATH/node_modules`.
+3. `$HOME/node_modules`.
 
 For each directory in search paths above:
 
--- a/docs/api/Module.md
+++ b/docs/api/Module.md
@@ -29,10 +29,10 @@ If a native module named `id` exists, load it and return.
 
 `require` function searches for modules in the following order:
 
-1. `$NODE_PRIORITIZED_PATH/node_modules` if present.
-1. `node_modules` folder under current working directory.
-2. `$NODE_PATH/node_modules`.
-3. `$HOME/node_modules`.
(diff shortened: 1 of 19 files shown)


yorkie:
This fixes #262, too.

legendecas:
IMHO, use `NODE_PRIORITIZED_PATH` environ to set a high priority search path rather than reversing search path order might be better

yorkie:
> IMHO, use NODE_PRIORITIZED_PATH environ to set a high priority search path rather than reversing search path order might be better

Neat! I will change this.

yorkie:
Added NODE_PRIORITIZED_PATH, @legendecas :)

docs/api/Module.md
```diff
@@ -29,10 +29,10 @@ If a native module named `id` exists, load it and return.
 
 `require` function searches for modules in the following order:
 
-1. Current working directory.
-2. `iotjs_modules` folder under current working directory.
-3. `$HOME/iotjs_modules`
-4. `$IOTJS_PATH/iotjs_modules`
+1. `$NODE_PRIORITIZED_PATH/node_modules` if present.
+1. `node_modules` folder under current working directory.
+2. `$NODE_PATH/node_modules`.
```
legendecas: Actually, if a `NODE_PATH` which is current working directory, relative js file can be loaded without leading `./`.

e.g. with following directory structure:

```
- /home/workspace
|- lib
|    |- foobar.js
|- index.js
```

and set `NODE_PATH` to `/home/workspace`,

and with following content in `index.js`, it works:

```js
require('lib/foobar')
```

So it should be `2. $NODE_PATH`

yorkie: The above example doesn't work at IoT.js and ShadowNode. The search path for NODE_PATH should be `$NODE_PATH/node_modules`.

legendecas: It works in Node.js, should we support it too?

yorkie:
Remove the node_modules postfix after NODE_PATH and NODE_PRIORITIZED_PATH @legendecas :)

PR #278, opened 2018-08-09
debug: add a new internal debug module for debugging based on env

yorkie:
##### Checklist

- [x] `npm test` passes

This is for logging informations for internal modules, for example:

```bash
$ NODE_DEBUG=module iotjs test.js
```

### DIFF
--- a/src/js/debug.js
+++ b/src/js/debug.js
@@ -1,3 +1,5 @@
+/* eslint strict: [2, "never"] */
+
 var NODE_DEBUG = process.env.NODE_DEBUG;
 
 module.exports = function(tag) {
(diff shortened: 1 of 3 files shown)


src/js/debug.js
```diff
@@ -0,0 +1,14 @@
+var NODE_DEBUG = process.env.NODE_DEBUG;
+
+module.exports = function(tag) {
+  function debug(str) {
+    var prefix = '\033';
+    console.info(`${tag} ${prefix}[90m${str}${prefix}[0m`);
```
legendecas: What about highlight tag, and leave message unmodified?

e.g. `${prefix}[90m${tag}[0m ${str}`

yorkie: Grey messages just don't make confusion on other normal logs.

PR #280, opened 2018-08-10
ci: clean up dependencies

legendecas:
##### Checklist

- [x] `npm test` passes

Removed docker service.

### DIFF
--- a/tools/travis_script.py
+++ b/tools/travis_script.py
@@ -9,6 +9,7 @@ from check_tidy import check_tidy
 
 platform = Platform()
 
+DOCKER_IMAGE = os.getenv('DOCKER_IMAGE', 'shadownode/base')
 DOCKER_ROOT_PATH = fs.join('/root')
 
 # ShadowNode path in travis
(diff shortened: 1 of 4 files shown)


tools/travis_script.py
```diff
@@ -34,12 +35,15 @@
...
 def run_docker():
     ex.check_run_cmd('docker', ['run', '-dit', '--privileged',
-                     '--name', DOCKER_NAME, '-v',
-                     '%s:%s' % (TRAVIS_BUILD_PATH, DOCKER_SHADOW_NODE_PATH),
-                     '--add-host', 'test.mosquitto.org:127.0.0.1',
-                     'iotjs/ubuntu:0.8'])
+                                '--name', DOCKER_NAME, '-v',
+                                '%s:%s' % (TRAVIS_BUILD_PATH,
+                                           DOCKER_SHADOW_NODE_PATH),
+                                '--add-host', 'test.mosquitto.org:127.0.0.1',
```
yorkie: We should add example.com here?

legendecas: There is no mock http/https server setup in container yet, so no host could be added.

yorkie:
Good job @legendecas, The [current CI Job](https://travis-ci.org/Rokid/ShadowNode/builds/414492872?utm_source=github_status&utm_medium=notification) only takes 9min.

PR #281, opened 2018-08-10
napi: object abstract operations

legendecas:
##### Checklist

- [x] `npm test` passes

N-API methods implemented in this PR:

- napi_create_error
- napi_create_type_error
- napi_create_range_error
- napi_instanceof
- napi_strict_equals

Fixes not clearing handled last thrown error.

### DIFF
--- a/src/internal/node_api_internal.h
+++ b/src/internal/node_api_internal.h
@@ -27,7 +27,7 @@
 
 #define NAPI_WEAK_ASSERT(error_t, assertion) \
   do {                                       \
-    if (!assertion)                          \
+    if (!(assertion))                        \
       return error_t;                        \
   } while (0)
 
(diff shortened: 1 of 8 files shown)


src/internal/node_api_internal.h
```diff
@@ -58,8 +58,8 @@ int napi_module_init_pending(jerry_value_t* exports);
 /** MARK: - node_api_env.c */
 napi_env iotjs_get_current_napi_env();
```
algebrait: A comment not on this PR.
Why the napi implementation depends on iot.js? Theoretically, only module loading is depended on iotjs, and async callling depends on libuv.

legendecas: The naming is used to separate N-API apis from shadow-node internal apis. Or shall we name it with prefix 'shadownode' or its abbreviation?

PR #284, opened 2018-08-13
module: implicit extension .node

### LINKED ISSUE #265: module: optional `.node` extension
* **Version**: 0.9.0
* **Platform**: all
* **Subsystem**: module

`require('build/binding')` could be used to load `build/binding.node` in node.js, which shall be supported in ShadowNode for compatibility.

legendecas:
##### Checklist

- [x] `npm test` passes

close #265

### DIFF
--- a/docs/api/Module.md
+++ b/docs/api/Module.md
@@ -38,6 +38,7 @@ For each directory in search paths above:
 
 - If a file `id` exists, load it and return.
 - If a file `id.js` exists, load it and retun.
+- If a file `id.node` exists, load it and retun.
 - If a directory `id` exists, module system consider the directory as a package:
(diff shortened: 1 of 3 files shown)


yorkie:
@legendecas The current implementation still makes many `fs.existsSync()` when implicitly loading the native module.

BTW, I shall not follow this feature, it causes the process bootstrap really that slow, except that we have a valuable reason.

legendecas:
Shall reconsider if implicit loading .node file is required in ShadowNode, closing for now

yorkie:
> Shall reconsider if implicit loading .node file is required in ShadowNode, closing for now

Yea, I think explicitly declare what type of the module you wanna use is always good way.

PR #286, opened 2018-08-13
package-manager: yarn

legendecas:
##### Checklist

- [ ] `npm test` passes


yorkie:
So the package-lock.json should be removed from source now :)

legendecas:
> So the package-lock.json should be removed from source now :)

Doesn't find package-lock.json in source file tree, is it added to git?

PR #293, opened 2018-08-14
assert: assert.throws expectation

### LINKED ISSUE #291: assert: assert.throws shall accept a regex matcher
* **Version**: 0.9.x
* **Platform**: all
* **Subsystem**: assert

```js
var assert = require('assert')

assert.throws(() => { throw new Error('foobar') }, /^foobar$/)
```

An assertion on regex matching on thrown error's message is commonly used in node.js. It has to be supported to adopt node.js test sets.

legendecas:
##### Checklist

- [x] `npm test` passes

close #291

### DIFF
--- a/src/js/assert.js
+++ b/src/js/assert.js
@@ -48,8 +48,8 @@ function AssertionError(options) {
   } else {
     this.message = getMessage(this);
   }
+  Error.captureStackTrace(this, AssertionError);
 }
-
 util.inherits(AssertionError, Error);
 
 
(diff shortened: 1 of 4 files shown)


src/js/assert.js
```diff
@@ -142,6 +139,48 @@ function doesNotThrow(block, message) {
...
+
+    // Handle primitives properly.
+    if (typeof actual !== 'object' || actual === null) {
+      var err = new AssertionError({
+        actual: actual,
+        expected: expected,
+        message: msg,
+        operator: 'throws',
+      });
+      throw err;
```
yorkie: No necessary to declare the new variable.

PR #300, opened 2018-08-16
napi: null terminator on copying string value

legendecas:
##### Checklist

- [x] `npm test` passes

Issues fixed in this PR:

- no null terminator added on end of string on copying string value
- returning number value `0` on N-API function would be transformed to `undefined`: now no tranformation is performed

### DIFF
--- a/src/napi/node_api_function.c
+++ b/src/napi/node_api_function.c
@@ -60,13 +60,11 @@ static jerry_value_t iotjs_napi_function_handler(
   }
 
   // TODO: check if nvalue_ret is escaped
-  if (nvalue_ret == NULL) {
-    jval_ret = jerry_create_undefined();
-  } else {
-    /** jval returned from N-API functions is scoped */
-    jval_ret = AS_JERRY_VALUE(nvalue_ret);
-  }
-
+  /**
+   * Do not turn NULL pointer into undefined since number value `0` in
+   * jerryscript also represented by NULL
+   */
+  jval_ret = AS_JERRY_VALUE(nvalue_ret);
 
 cleanup:
   jerryx_remove_handle(scope, jval_ret, &jval_ret);
--- a/src/napi/node_api_value.c
+++ b/src/napi/node_api_value.c
@@ -202,8 +202,8 @@ napi_status napi_get_value_string_utf8(napi_env env, napi_value value,
   jerry_value_t jval = AS_JERRY_VALUE(value);
   NAPI_TRY_TYPE(string, jval);
 
+  size_t str_size = jerry_get_utf8_string_size(jval);
   if (buf == NULL) {
-    size_t str_size = jerry_get_utf8_string_size(jval);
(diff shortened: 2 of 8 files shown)


src/napi/node_api_value.c
```diff
@@ -204,12 +204,15 @@ napi_status napi_get_value_string_utf8(napi_env env, napi_value value,
...
   if (buf == NULL) {
     size_t str_size = jerry_get_utf8_string_size(jval);
+    /** null terminator is excluded */
     NAPI_ASSIGN(result, str_size);
     NAPI_RETURN(napi_ok);
   }
 
   jerry_size_t written_size =
       jerry_string_to_utf8_char_buffer(jval, (jerry_char_t*)buf, bufsize);
+  /** expects one more byte to write null terminator  */
```
yorkie: Ditto.

test/napi-testsets.json
```diff
@@ -55,6 +55,7 @@
     { "name": "test_uv_loop/test.js", "skip": ["all"], "reason": "uv api not implemented" }
   ],
   "napi": [
-    { "name": "napi.test.js" }
+    { "name": "napi_string.test.js", "skip": false },
+    { "name": "napi.test.js", "skip": false }
```
yorkie: `skip` should be by default?

src/napi/node_api_function.c
```diff
@@ -60,13 +60,11 @@ static jerry_value_t iotjs_napi_function_handler(
...
-    jval_ret = jerry_create_undefined();
-  } else {
-    /** jval returned from N-API functions is scoped */
-    jval_ret = AS_JERRY_VALUE(nvalue_ret);
-  }
-
+  /**
+   * Do not turn NULL pointer into undefined since number value `0` in
+   * jerryscript also represented by NULL
+   */
```
yorkie: How can we handle those test cases to assert on zero?

test/napi-testsets.json
legendecas: No, it will skip non-builtin module by default 😂

src/napi/node_api_value.c
```diff
@@ -204,12 +204,15 @@ napi_status napi_get_value_string_utf8(napi_env env, napi_value value,
 
   if (buf == NULL) {
     size_t str_size = jerry_get_utf8_string_size(jval);
+    /** null terminator is excluded */
```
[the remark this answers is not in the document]
legendecas: `/**` can be turned into multi-line comment with ease :)

```
/**
 * Like this
 */
```

src/napi/node_api_function.c
legendecas: String length tests covered this part.

src/napi/node_api_value.c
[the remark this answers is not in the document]
yorkie: This is a single line, use `/* */` or `//` instead, please :)

PR #306, opened 2018-08-18
fix: napi_wrap shall be weak reference

legendecas:
##### Checklist

- [x] `npm test` passes

Issues addressed in this PR:

- napi_wrap shall return a weak reference;
- keep napi_new_instance aligned with napi_call_function.


yorkie:
Seems some style problems :)

test/napi/napi_object_wrap.test.js
```diff
@@ -0,0 +1,12 @@
+var assert = require('assert');
```
yorkie: use strict.

PR #311, opened 2018-08-19
console: fix print object for single parameter

ymrdf:
Format console.log's param using util.formatValue when the function has only one param. The performance is the same as before:
```
console/common.js n=128: 6,353.840618673461   #before
console/common.js n=128: 6,344.13890531784     #after
```

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/js/console.js
+++ b/src/js/console.js
@@ -48,7 +48,7 @@ function stderr(text) {
 Console.prototype.log =
 Console.prototype.debug =
 Console.prototype.info = function() {
-    stdout(util.format.apply(this, arguments) + '\n');
+  stdout(util.format.apply(this, arguments) + '\n');
 };
 
 Console.prototype.warn =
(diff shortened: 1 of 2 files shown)


src/js/console.js
```diff
@@ -49,7 +49,7 @@ Console.prototype.log =
 Console.prototype.debug =
 Console.prototype.info = function() {
   if (arguments.length === 1) {
-    stdout(arguments[0] + '\n');
+    stdout(util.formatValue(arguments[0]) + '\n');
```
yorkie: use StringTemplate please.

src/js/console.js
```diff
@@ -58,7 +58,7 @@ Console.prototype.info = function() {
 Console.prototype.warn =
 Console.prototype.error = function() {
   if (arguments.length === 1) {
-    stderr(arguments[0] + '\n');
+    stderr(util.formatValue(arguments[0]) + '\n');
```
yorkie: Ditto.

yorkie:
And the common/console is only testing the empty string, see https://github.com/Rokid/ShadowNode/blob/master/benchmark/console/common.js#L12.

yorkie:
Thanks @ymrdf, now this looks good to me.

PR #315, opened 2018-08-20
leak: fix memory leak introduced in #272

algebrait:
##### Checklist

- [x] `npm test` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "lint": "tools/check_tidy.py && eslint .",
     "test": "npm run lint && npm run test-jerry && npm run test-iotjs",
     "test-jerry": "deps/jerry/tools/run-tests.py --unittests",
-    "test-iotjs": "tools/build.py --run-test",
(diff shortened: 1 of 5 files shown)


algebrait:
If open valgrind, ci time will increase.

package.json
```diff
@@ -6,7 +6,7 @@
     "lint": "tools/check_tidy.py && eslint .",
     "test": "npm run lint && npm run test-jerry && npm run test-iotjs",
     "test-jerry": "deps/jerry/tools/run-tests.py --unittests",
-    "test-iotjs": "tools/build.py --run-test --no-check-valgrind",
+    "test-iotjs": "tools/build.py --run-test",
```
yorkie: @algebrait CI doesn't use this to enable/disable valgrind, and I think we might try to open it at [this line](https://github.com/Rokid/ShadowNode/blob/master/tools/travis_script.py#L27)?

algebrait:
travis problem is tracked by #319

yorkie:
Thanks for this fix @algebrait :)

PR #316, opened 2018-08-20
dbus: expose the internal member privately

yorkie:
##### Checklist

- [ ] `npm test` passes

Exposes the low-level member `_bus` for signal handler usage.


yorkie:
Thanks @legendecas, just created _bus as an non-enumerable value.

PR #318, opened 2018-08-20
add util.promisify

legendecas:
##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/js/util.js
+++ b/src/js/util.js
@@ -250,6 +250,34 @@ function deprecate(fn, msg, code) {
   return deprecated;
 }
 
+// FIXME: use Symbol on implementation done
+var customPromisifySymbol = 'util:promisify:custom';
+promisify.custom = customPromisifySymbol;
+function promisify(original) {
+  if (typeof original != 'function') {
+    throw new TypeError('expect a function on promisify');
+  }
+  if (typeof original[customPromisifySymbol] == 'function') {
+    return original[customPromisifySymbol];
+  }
+  return function promisified() {
+    var args = Array.prototype.slice.call(arguments, 0);
+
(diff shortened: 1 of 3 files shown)


src/js/util.js
```diff
@@ -250,6 +250,33 @@ function deprecate(fn, msg, code) {
...
+  }
+  return function promisified() {
+    var args = Array.prototype.slice.call(arguments, 0);
+
+    return new Promise((resolve, reject) => {
+      args.push(callback);
+      original.apply(this, args);
+
+      function callback(err, result) {
+        if (err != null) {
```
yorkie: Should we verify this is an `Error` instance?

src/js/util.js
```diff
@@ -250,6 +250,33 @@ function deprecate(fn, msg, code) {
   return deprecated;
 }
 
+var customPromisifySymbol = 'util:promisify:custom';
```
yorkie: This looks like the symbol, so i think it's better to flag a TODO or FIXME as the following:

```js
// FIXME(legendecas): change to use formal `Symbol` when it is ready.
```

src/js/util.js
legendecas: No

yorkie: I saw the same function on Node.js looks like this:

```js
if (err)
  return reject(err)
```

Why do we make this difference here?

legendecas: `if (err)` would be true if err is truthy, `err != null` can only be true if err is `null` or `undefined`.

yorkie: Yay! How about fixing it in Node.js core, either?

yorkie:
@legendecas Looks good to me, please add document as well :)

PR #322, opened 2018-08-20
child_process: add tests for env

yorkie:
##### Checklist

- [x] `npm test` passes

Just added the test case for env setting.

### DIFF
--- a/test/node/parallel/test-child-process-env.js
+++ b/test/node/parallel/test-child-process-env.js
@@ -35,14 +35,7 @@ Object.setPrototypeOf(env, {
   'FOO': 'BAR'
 });
 
-var child;
-if (false) {
-  child = spawn('cmd.exe', ['/c', 'set'],
-                Object.assign({}, process.env, { env: env }));
-} else {
-  child = spawn('/usr/bin/env', [],
-                Object.assign({}, process.env, { env: env }));
-}
+var child = spawn('/usr/bin/env', [], Object.assign({}, process.env, env));
 
 var response = '';
 child.stdout.setEncoding('utf8');
--- a/test/node/parallel/test-child-process-env.js
+++ b/test/node/parallel/test-child-process-env.js
@@ -1,0 +1,61 @@
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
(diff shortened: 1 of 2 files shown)


test/node/parallel/test-child-process-env.js
```diff
@@ -0,0 +1,61 @@
...
+  'EMPTY': ''
+};
+Object.setPrototypeOf(env, {
+  'FOO': 'BAR'
+});
+
+var child;
+if (false) {
+  child = spawn('cmd.exe', ['/c', 'set'],
+                Object.assign({}, process.env, { env: env }));
```
legendecas: Do you mean `{ env: Object.assign({}, process.env, env) }`?

test/node/parallel/test-child-process-env.js
```diff
@@ -0,0 +1,61 @@
...
+} else {
+  child = spawn('/usr/bin/env', [],
+                Object.assign({}, process.env, { env: env }));
```
legendecas: Ditto

test/node/parallel/test-child-process-env.js
```diff
@@ -0,0 +1,61 @@
...
+  'UNDEFINED': undefined,
+  'NULL': null,
+  'EMPTY': ''
+};
+Object.setPrototypeOf(env, {
+  'FOO': 'BAR'
+});
+
+var child;
+if (false) {
```
legendecas: What does `if (false)` mean here?

test/node/parallel/test-child-process-env.js
```diff
@@ -0,0 +1,61 @@
...
+});
+
+process.on('exit', function() {
+  assert.ok(response.search('HELLO=WORLD') > -1);
+  assert.ok(response.search('FOO=BAR') > -1);
+  assert.ok(response.search('UNDEFINED=undefined') > -1);
+  assert.ok(response.search('NULL=null') > -1);
+  assert.ok(response.search(`EMPTY=${os.EOL}`) > -1);
+  console.log('done');
+});
```
legendecas: missing EOF

test/node/parallel/test-child-process-env.js
yorkie: This is copied from Node.js, previously it's for windows.

yorkie:
Close because of inactivity.

PR #323, opened 2018-08-21
dbus: refactor and add start adding tests

yorkie:
##### Checklist

- [x] `npm test` passes

This PR is going to refactor the D-BUS module which is lacks of the unit tests. And it also fixes some memory leaks and improves something better:

- Moves `sax` in need, which is required only by introspect feature.
- Uses the single connection between service and `getBus()`.
- Frees the some acquired callbacks when no need to use.
- Frees the string when decoding messages.
- Fixes the message reply unavailable by referencing the `DBusMessage*`.


yorkie:
@legendecas @algebrait @LanFly The CI seems green already in my last commit, could you please take a review now?

src/js/dbus.js
```diff
@@ -527,10 +541,16 @@ function registerService(name, service) {
 exports.Define = function(type) {
   if (type === String) {
     return 's';
-  } else if (type === 'Number') {
-    return 'i';
+  } else if (type === Number) {
```
legendecas: maybe use map style?
```js
{
  Number: 'i',
  String: 's',
...
}
```

test/run_pass/test_dbus_service_and_call.js
```diff
@@ -0,0 +1,45 @@
...
+  'test.dbus.myservice.Interface1', 'testWithArgs', 's', ['foobar'], (err, res) => {
+  assert.equal(err, null);
+  assert.equal(res, 'foobar!!!');
+  plan += 1;
+});
+
+setTimeout(function() {
+  assert.equal(plan, 2);
+  bus.destroy();
+}, 500);
```
legendecas: missing EOF

src/js/dbus.js
yorkie: I tried this, but JerryScript seems not support this way.

src/js/dbus.js
```diff
@@ -463,7 +472,7 @@ ServiceInterface.prototype.addMethod = function(name, opts, handler) {
  * @method addProperty
  */
 ServiceInterface.prototype.addProperty = function() {
-  // TODO no supported
+  throw new Error('addProperty not implemented');
```
legendecas: error shall be fatal, and could not be catched

src/js/dbus.js
```diff
@@ -496,7 +505,7 @@ ServiceInterface.prototype.emit = function(name, val) {
  * @method update
  */
 ServiceInterface.prototype.update = function() {
-  // Do nothing, just for comp.
+  console.error('update is deprecated, please remove');
```
legendecas: use `util.deprecate`

yorkie:
Fixed @legendecas :)

test/testsets.json
```diff
@@ -11,6 +11,27 @@
     { "name": "test_console.js" },
     { "name": "test_crypto.js" },
     { "name": "test_crypto_hash.js" },
+    { "name": "test_dbus_after_destroy.js",
+      "env": {
+        "DISPLAY": ":0"
```
legendecas: What does "DISPLAY" mean?

tools/testrunner.py
```diff
@@ -244,7 +244,7 @@ def run_test(self, testfile, options):
         my_env = os.environ.copy()
         if options["env"] != None:
             for key, val in options["env"].items():
-                if key == u"NODE_PATH":
+                if key == u"NODE_PATH" or key == u"DISPLAY":
```
legendecas: should `DISPLAY` be expanded here?

PR #333, opened 2018-09-14
 debugger: Error.stack deal with top level source

algebrait:
##### Checklist

- [x] `npm test` passes
test/run_fail/test_iotjs_type_error.js

### DIFF
--- a/src/js/iotjs.js
+++ b/src/js/iotjs.js
@@ -76,13 +76,13 @@
       if (/.*:/.test(line)) {
         file = line.slice(0, -1);
       } else {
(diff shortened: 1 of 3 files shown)


src/js/iotjs.js
```diff
@@ -76,13 +76,13 @@
       if (/.*:/.test(line)) {
         file = line.slice(0, -1);
       } else {
-        var m = line.match(/(\+ ([a-zA-Z0-9_]*))? \[(\d+),(\d+)\] (\d+)/);
+        var m = line.match(/(\+ ([a-zA-Z0-9_]*))?( \[(\d+),(\d+)\])? (\d+)/);
```
algebrait: Top level source has no function name. Line in dump file is just like
```
 12345
```
So [line,column] should be optinal.

PR #339, opened 2018-09-18
fix(n-api): asserts on execution thread

legendecas:
##### Checklist

- [x] `npm test` passes


yorkie:
LGTM if add a failure test under test/run_fail :)

PR #348, opened 2018-09-23
http: fix memory leak when socket dns lookup error

lolBig:
- leak fix
- remove dns loop up error, because the error will be emitted twice


test/run_pass/test_net_http_dns_error.js
```diff
@@ -0,0 +1,18 @@
...
+  process.nextTick(() => {
+    var destroyed = req.socket._socketState.destroyed;
+    console.log(`socket is destroyed ${host}:`, destroyed)
+    assert.equal(destroyed, true)
+  })
+})
+// handle error event to prevent the process throw an error
+req.on('error', (err) => {
+  console.log('request error', err.message)
+})
```
legendecas: Missing EOF :P

test/run_pass/test_net_http_dns_error.js
```diff
@@ -0,0 +1,18 @@
...
+req.socket.on('finish', () => {
+  process.nextTick(() => {
+    var destroyed = req.socket._socketState.destroyed;
+    console.log(`socket is destroyed ${host}:`, destroyed)
+    assert.equal(destroyed, true)
+  })
+})
+// handle error event to prevent the process throw an error
+req.on('error', (err) => {
+  console.log('request error', err.message)
```
legendecas: Maybe an assertion such as `mustCall` could be added here instead?

test/run_pass/test_net_http_dns_error.js
```diff
@@ -0,0 +1,18 @@
+var assert = require('assert');
+var http = require('http');
+
+var host = 'www.unable-to-resolved-host.com'
+var req = http.request({
+  host: host,
+})
+req.socket.on('finish', () => {
+  process.nextTick(() => {
```
legendecas: Assertions such as `mustCall` could be used here

PR #352, opened 2018-09-25
fix leak

lolBig:


src/js/https_client.js
```diff
@@ -218,7 +218,11 @@ function parserOnIncomingClient(res, shouldKeepAlive) {
 function responseOnEnd() {
   var res = this;
   var req = res.req;
-  req.socket.end();
+  var socket = req.socket;
+
+  if (socket._socketState.writable) {
+    socket.destroySoon();
```
legendecas: what about `socket#destroy` instead of `socket#destroySoon` since `socket#destroy ` would safely skip if socket was destroyed already.

yorkie: @legendecas The [http_client.js](https://github.com/Rokid/ShadowNode/blob/master/src/js/https_client.js) is using `destroySoon`, my suggestion is keep consistent between http & https to fix this issue. Or we need to change both these two.

PR #357, opened 2018-09-29
events: Emit remove listener when remove listener.

Txiaozhe:
##### Checklist

- [ ] `npm test` passes


benchmark/events/events-newListener.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
```
legendecas: I don't think this file shall be included in this PR. Please try to remove this file :p

src/js/events.js
```diff
@@ -67,7 +67,9 @@ EventEmitter.prototype.addListener = function(type, listener) {
     throw new TypeError('listener must be a function');
   }
 
-  this.emit('newListener', type, listener);
```
legendecas: Ditto

benchmark/events/events-newListener.js
Txiaozhe: I am sorry, I confuse the two pr, I will fix it. Thanks!

PR #359, opened 2018-10-01
dbus: support returning double type

yorkie:
##### Checklist

- [x] `npm test` passes

This supports the `double` type for dbus module.

### DIFF
--- a/test/run_pass/test_dbus_service_and_call.js
+++ b/test/run_pass/test_dbus_service_and_call.js
@@ -16,14 +16,24 @@ myiface.addMethod('testWithArgs', {
   out: [dbus.Define(String)] }, (num, cb) => {
   cb(null, num + '!!!');
 });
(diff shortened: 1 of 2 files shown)


test/run_pass/test_dbus_service_and_call.js
```diff
@@ -16,31 +16,45 @@ myiface.addMethod('testWithArgs', {
...
   assert.equal(res, 'foobar!!!');
   plan += 1;
 });
 
+bus.callMethod(
+  'org.myservice',
+  '/org/myobject',
+  'test.dbus.myservice.Interface1', 'testWithNumber', 's', [], (err, res) => {
+  assert.equal(err, null);
+  assert.equal(res, 9999);
```
legendecas: Use float pointer value for test might be more precise for this PR 😄

PR #364, opened 2018-10-09
events: Limit listener number

Txiaozhe:
##### Checklist

- [x] `npm test` passes

limit listener number like in node, avoid the listener leak risk.


src/js/events.js
```diff
@@ -162,9 +201,24 @@ EventEmitter.prototype.listeners = function(type) {
 };
 
 
-EventEmitter.prototype.setMaxListeners = function() {
-  // EMPTY
-  // This is for compatible with Node.js API, but no effects inside here.
+EventEmitter.prototype.setMaxListeners = function(n) {
+  if (typeof n !== 'number' || n < 0) {
+    throw new Error('arguments of setMaxListeners must be a ' +
```
yorkie: This might be a `TypeError`.

src/js/events.js
```diff
@@ -162,9 +201,24 @@ EventEmitter.prototype.listeners = function(type) {
...
+    'non-negative number');
+  }
+  this._maxListeners = n;
+  return this;
+};
+
+function $getMaxListeners(that) {
+  if (that._maxListeners === undefined) {
+    return EventEmitter.defaultMaxListeners;
```
yorkie: Where the `EventEmitter.defaultMaxListeners` was defined?

src/js/events.js
```diff
@@ -24,12 +24,31 @@ function EventEmitter() {
 module.exports = EventEmitter;
 module.exports.EventEmitter = EventEmitter;
 
+EventEmitter.prototype._eventsCount = 0;
+EventEmitter.prototype._maxListeners = undefined;
+
+var defaultMaxListeners = 10;
```
yorkie: Should the default maximum number to be 10 in embeddable device for ShadowNode?

src/js/events.js
```diff
@@ -24,12 +24,31 @@ function EventEmitter() {
...
+
+EventEmitter.prototype.setDefaultMaxListeners = function(n) {
+  if (typeof n !== 'number' || n < 0 || Number.isNaN(n)) {
+    throw new Error('defaultMaxListeners must be a non-negative number');
```
yorkie: `TypeError`, too.

src/js/events.js
Txiaozhe: I know less on embeddable device, but how much the default maximum number I should set, more or less? Or no need do this?
In my submission, the hardware of embeddable device is short board, if we set too more listeners, this will lead to memory failure. So we should set a smaller value. Like 8 or 5???

docs/api/Events.md
```diff
@@ -222,6 +222,28 @@ res = emitter.emit('event3'); // res == false
...
+* Returns `emitter` {events.EventEmitter}.
+
+Set the max listener limit, must be number and > 0.
+
+### emitter.getMaxListeners()
+* Returns {number} The max listener limit.
+
+Get the max listener limit, must be number.
+
+### emitter.setDefaultMaxListeners(num)
```
legendecas: I'm wondering if `setDefaultMaxListeners` and `getDefaultMaxListeners` shall be an instance method rather than a static method, yet since EventEmitter.defaultMaxListeners could be defined as an accessor, both `setDefaultMaxListeners` and `getDefaultMaxListeners` are redundant in my opinion.

Txiaozhe: Exactly, I will define an accessor. Thanks!

src/js/events.js
Txiaozhe: Although I don't think there is not big difference between 7 and 10, it is also a good choice. @(｡･o･)@

docs/api/Events.md
```diff
@@ -222,6 +222,28 @@ res = emitter.emit('event3'); // res == false
 
 ```
 
+### emitter.setMaxListeners(num)
+* `num` {number} The max listener limit, default 10.
```
yorkie: The documentation is required to update, too.

docs/api/Events.md
```diff
@@ -222,6 +222,28 @@ res = emitter.emit('event3'); // res == false
...
+* Returns `emitter` {events.EventEmitter}.
+
+Set the max listener limit, must be number and > 0.
```
yorkie: Must be a number and > 0

docs/api/Events.md
```diff
@@ -222,6 +222,28 @@ res = emitter.emit('event3'); // res == false
...
+
+### emitter.getMaxListeners()
+* Returns {number} The max listener limit.
+
+Get the max listener limit, must be number.
+
+### emitter.setDefaultMaxListeners(num)
+* `num` {number} The default max listener limit, default 10.
```
yorkie: Ditto.

Txiaozhe: awkward >__<|||, thanks!

docs/api/Events.md
```diff
@@ -222,6 +222,17 @@ res = emitter.emit('event3'); // res == false
 
 ```
 
+### emitter.setMaxListeners(num)
+* `num` {number} The max listener limit, default 7.
```
yorkie: use Number

Txiaozhe:
I'm going to come to an end to 'events' revamp, and I'm going to start revamping other apis, and if you have specific requirements, you can tell me, I'm going to do. Thanks! @yorkie @legendecas and others.

test/common/index.js
```diff
@@ -45,3 +45,21 @@ got ${it.actual}
 ${it.stack}`);
   });
 }
+
+function expectsError(fn, exact) {
+  function innerFn(error) {
```
legendecas: I'm wondering if `assert.throws` works exactly the same with this function?

Txiaozhe: Yes, because I wanted to do a more perfection assert in this function just like in Node, but there are some operations that are not supported in ShadowNode, so I reserve it. ：)
Next I will revamp that api, I will do it!

Txiaozhe: Thanks!

PR #366, opened 2018-10-10
leak: fix http/https and os leak

lolBig:
1. when the network is disconnected,  the socket cloud not been destroyed if the HTTP/HTTPS request through ip
2. invoke uv_free_interface_addresses to free uv_interface_address_t*
3. remove the 'lookup' event listener in https

### DIFF
--- a/test/node/parallel/test-http-abort.js
+++ b/test/node/parallel/test-http-abort.js
@@ -12,10 +12,10 @@ function getHandle(url) {
   throw new Error('unsupported url ' + url);
 }
 
-// // test http_client
-test('http://www.example.com/');
-// // test https_client
-test('https://www.example.com/');
+// test http_client
+test('http://example.com/');
+// test https_client
+test('https://example.com/');
 // unreachable tunnels, manually abort them
 var req1 = test('http://127.0.0.2');
 setTimeout(function () {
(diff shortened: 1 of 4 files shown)


test/node/parallel/test-http-abort.js
```diff
@@ -11,10 +12,19 @@ function getHandle(url) {
   throw new Error('unsupported url ' + url);
 }
 
-// test http_client
-test('http://example.com/');
-// test https_client
-test('https://example.com/');
+// // test http_client
```
legendecas: duplicated double slash?

test/node/parallel/test-http-abort.js
```diff
@@ -23,20 +33,27 @@ function test(url) {
   var handle = getHandle(url);
   var req = handle.get(url, function(res) {
     res.on('data', function(chunk) {
-      assert.strictEqual(isAborted, false, 'should not aborted');
+      console.log(`${url} ondata`);
```
legendecas: unexpected console log

test/node/parallel/test-http-abort.js
```diff
@@ -23,20 +33,24 @@ function test(url) {
...
   });
 
-  req.on('abort', function() {
+  req.on('abort', common.mustCall(function() {
     eventTriggered = true;
-  });
+  }));
+
+  req.socket.on('close', common.mustCall(function() {
+  }));
```
yorkie: Why handle the close on the empty function?

lolBig: to ensure that "close" events can be triggered, that is, the socket can be destroyed

PR #368, opened 2018-10-10
jerry: ref to function self should not create new object

algebrait:
##### Checklist

- [x] `npm test` passes
       test/run_pass/test_iotjs_selfref.js

### DIFF
--- a/test/run_pass/test_iotjs_selfref.js
+++ b/test/run_pass/test_iotjs_selfref.js
@@ -1,0 +1,19 @@
+/* Copyright 2017-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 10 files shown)


test/run_pass/test_iotjs_selfref.js
```diff
@@ -0,0 +1,19 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+var assert = require('assert');
+
+var foo = function bar() {
+  assert(foo === bar)
+}
```
yorkie:
```js
var foo = function bar() {
  assert(foo === bar);
};
```
:)

PR #370, opened 2018-10-15
mqtt: clean up the keepalive timer when disconnecting

yorkie:
##### Checklist

- [x] `npm test` passes

The timer for *Keep Alive* should be cleared if we explicitly disconnect it.

/cc @lolBig @legendecas

### DIFF
--- a/src/js/mqtt.js
+++ b/src/js/mqtt.js
@@ -74,8 +74,8 @@ MqttClient.prototype.connect = function() {
     this._socket = net.connect(opts, this._onconnect.bind(this));
   }
   this._socket.on('data', this._ondata.bind(this));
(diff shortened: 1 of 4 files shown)


src/js/mqtt.js
```diff
@@ -97,19 +97,16 @@ MqttClient.prototype._onconnect = function() {
...
+  this.emit('error', err);
+  this._ondisconnect();
 };
 
 MqttClient.prototype._onend = function() {
+  this._clearKeepAlive();
   this._ondisconnect();
 };
 
 MqttClient.prototype._ondisconnect = function(err) {
```
lolBig: remove useless argument 'error'

PR #372, opened 2018-10-15
process: add signal events support

yorkie:
##### Checklist

- [x] `npm test` passes

Added the signal events support by Node.js API, and this is mainly the precondition of simplifying the progress of taking the snapshot, and other debugger/dump usages.

/cc @lolBig @legendecas @fighterkin @algebrait

### DIFF
--- a/deps/libtuv/src/unix/core.c
+++ b/deps/libtuv/src/unix/core.c
@@ -154,6 +154,12 @@ void uv_close(uv_handle_t* handle, uv_close_cb close_cb) {
     uv__poll_close((uv_poll_t*)handle);
     break;
 
+  case UV_SIGNAL:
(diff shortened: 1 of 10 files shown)


legendecas:
Does invocation of `process.exit` in signal listeners work immediately if a timer still waiting?

yorkie:
The Node.js documentation about `process.exit`:

> Calling `process.exit()` will force the process to exit as quickly as possible even if there are still asynchronous operations pending that have not yet completed fully, including I/O operations to process.stdout and process.stderr.

So the answer should be yes, `process.exit()` exits the current process immediately.

yorkie:
@legendecas Added an issue #373 about the current wrong at `process.exit()`, and we will address that in another patch.

lolBig:
LGTM

PR #375, opened 2018-10-17
TLS:  fix tls connection could not be actively disconnected

lolBig:


test/run_pass/test_mqtts_client.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+var testHost = 'mqtts://test.mosquitto.org:8883';
+var mqtt = require('mqtt');
+var assert = require('assert');
+var common = require('../common')
```
yorkie:
```suggestion
var common = require('../common');
```

PR #376, opened 2018-10-17
websocket: fixed crash due to HTTP exception

lolBig:


src/js/websocket.js
```diff
@@ -210,6 +210,9 @@ WebSocketClient.prototype.handshake = function(location, subProtocol) {
       self.connected = true;
     }
   });
+  httpConnection.on('error', function (err) {
+    self.emit('error', err)
```
yorkie:
```suggestion
    self.emit('error', err);
```

src/js/websocket.js
```diff
@@ -210,6 +210,9 @@ WebSocketClient.prototype.handshake = function(location, subProtocol) {
       self.connected = true;
     }
   });
+  httpConnection.on('error', function (err) {
```
yorkie:
```suggestion
  httpConnection.on('error', function(err) {
```

PR #240, opened 2018-07-31
jerry: add heap profiler

algebrait:
##### Checklist

- [x] `npm test` passes
- [x] tests and/or benchmarks are included
  - test/run_pass/test_profilerjs
- [x] documentation is changed or added
  - docs/api/Profiler.md
  - docs/devs/Optimization-Tips.md
- [x] object node
- [x] string node
- [x] number node
- [x] bytecode node
- [x] synthetic node
- [x] edge in mark-sweep gc's reference

PR #379, opened 2018-10-19
process: add shebang parser supports

yorkie:
##### Checklist

- [x] `npm test` passes(failed on an unrelated suite)

This adds the shebang feature that makes the script could be an executable file. BTW, this adds the jerry tests into our CI tests set, too.

And now it fails on [deps/jerry/tests/jerry/regression-test-issue-2105.js](https://github.com/Rokid/ShadowNode/blob/master/deps/jerry/tests/jerry/regression-test-issue-2105.js):

```js
/* Check properties of global object */
assert(Object.keys(this) == "assert,gc,print,a,fail,fail_two");
```

@algebrait The actual this in global is: `assert,gc,print,heapdump,a,fail,fail_two`, so I guess #240 should be relevant with this issue.

Before merging this, we have to resolve the jerry tests failure at first.


yorkie:
Thanks @algebrait, I have moved the implementation into Node.js runtime, and let's revisit the jerry tests in another PR :)

test/run_fail/test_shebang_fail.js
```diff
@@ -0,0 +1,4 @@
+
+#!/usr/bin/env fail
+
+// this should be fail
```
legendecas: missing eof :)

legendecas: linter has been disabled in our test directory, which is not very convenient to detect code flaws and auto-format with linter. It has to be reconsidered how we apply the lint to the tests.

PR #380, opened 2018-10-20
async-work: fix complete callback segv/ticks

legendecas:
##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/napi/node_api_async.c
+++ b/src/napi/node_api_async.c
@@ -57,6 +57,8 @@ static void iotjs_uv_work_after_cb(uv_work_t* req, int status) {
       jerry_release_value(jval_err);
     }
   }
+
+  iotjs_process_next_tick();
 }
 
 napi_status napi_create_async_work(napi_env env, napi_value async_resource,
--- a/src/napi/node_api_async.c
+++ b/src/napi/node_api_async.c
@@ -98,7 +98,7 @@ napi_status napi_delete_async_work(napi_env env, napi_async_work work) {
   NAPI_TRY_ENV(env);
   uv_work_t* work_req = (uv_work_t*)work;
   iotjs_async_work_t* async_work = (iotjs_async_work_t*)work_req->data;
-  free(async_work);
+  IOTJS_RELEASE(async_work);
   NAPI_RETURN(napi_ok);
 }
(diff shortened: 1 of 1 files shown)


src/napi/node_api_async.c
```diff
@@ -57,6 +61,8 @@ static void iotjs_uv_work_after_cb(uv_work_t* req, int status) {
       jerry_release_value(jval_err);
     }
   }
+
+  iotjs_process_next_tick();
```
yorkie: Please add a regression on https://github.com/Rokid/ShadowNode/blob/1a7a7649be26330cebde42131808142eaede5b3b/test/napi/napi_async.test.js :)

yorkie:
Ah, we did skip the test_async/test.js at https://github.com/Rokid/ShadowNode/blob/1a7a7649be26330cebde42131808142eaede5b3b/test/napi-testsets.json#L15

@legendecas Could you please remove the skip flag now?

src/napi/node_api_async.c
legendecas: tests are already included https://github.com/Rokid/ShadowNode/blob/1a7a7649be26330cebde42131808142eaede5b3b/test/napi/napi_async.test.js#L12.
Yet it doesn't test as expected though.

legendecas:
@yorkie it's not possible to enable `test_async/test.js` pulled from node repo for not implemented sync api
 https://github.com/Rokid/ShadowNode/blob/master/test/addons-napi/test_async/test.js#L17

yorkie:
Or could we have the patches to N-API Test Suites just like the openwrt/buildroot package, which provides a way to cover more NTS without any forks with source?

@mhdawson @gabrielschulhof, I think this tooling could also be included into our future NTS repository, and it helps implementor to get more coverages.

yorkie:
@legendecas we should avoid any case that uses freed block, so please set the pointer to be null after it gets freed at: https://github.com/Rokid/ShadowNode/blob/c56f4f96bd53259befece9f2ea996fe784f98a7e/src/napi/node_api_async.c#L101

And use `IOTJS_ALLOC` and `IOTJS_RELEASE`, too.

Issue #294, opened 2018-08-14
jerry: external function `this` is not instance of target


legendecas:
* **Version**: 0.9.x
* **Platform**: all
* **Subsystem**: jerryscript

Short brief:

While constructing an external function, argument `this` is not instance of the external function.

A detailed example can be found in #295 ([goto](https://github.com/Rokid/ShadowNode/pull/295/files#diff-24c74757a99c1d0a58827b53177ba016R14))

legendecas:
Issue caused by https://github.com/Rokid/ShadowNode/blob/master/deps/jerry/jerry-core/api/jerry.c#L2413, which defines this as an undefined.

yorkie:
This should be fixed at #386.

PR #381, opened 2018-10-20
jerry: merge a59cf4f from upstream to optimize the function call

yorkie:
This merges the upstream PR(https://github.com/jerryscript-project/jerryscript/pull/2414) which optimizes the following:

> Furthermore add a construct flag, which disallows calling certain functions without new. Constructing bound arrow functions correctly throws error now.

##### Checklist

- [x] `npm test` passes

BTW, https://github.com/jerryscript-project/jerryscript/commit/65ae949dc32ebf8c06df8f5e0fd4441b1bef5009 is valuable to merge into our own fork, either.


yorkie:
Because the current jerry tests are fails, so just run test locally with the following result:

> [64/     369] build/tests/jerry_tests-es2015_subset-debug/bin/jerry  tests/jerry/es2015/regression-test-issue-2414.js: PASS

legendecas:
Looks like #294 doesn't get resolved in this function call rework?

yorkie:
> Looks like #294 doesn't get resolved in this function call rework?

This doesn't fixes that, but I have fixed at #386.

PR #383, opened 2018-10-21
promise: ShadowNode use seperate profile instead of polluting jerry

algebrait:
##### Checklist

- [x] `npm test` passes


algebrait:
This PR will pass promise cases in jerry-tests

CMakeLists.txt
```diff
@@ -55,7 +55,7 @@ if(NOT DEFINED ENABLE_LTO)
 endif()
 
 if(NOT DEFINED DISABLE_ES2015)
-  set(FEATURE_PROFILE "es2015-subset")
+  set(FEATURE_PROFILE "ShadowNode")
 endif()
```
yorkie: I think we can removed `DISABLE_ES2015` and use the `shadow-node` directly or by default.

deps/jerry/jerry-core/profiles/ShadowNode.profile
```diff
@@ -0,0 +1,2 @@
+CONFIG_DISABLE_ES2015_PROMISE_BUILTIN
```
yorkie: To keep the consistent with other profiles, suggestion rename: `shadow-node.profile`.

BTW, could you please also move the this profile to ShadowNode root directory, JerryScript supports the absolute path at [deps/jerry/jerry-core/CMakeLists.txt#L236-L247](https://github.com/Rokid/ShadowNode/blob/master/deps/jerry/jerry-core/CMakeLists.txt#L236-L247).

CMakeLists.txt
```diff
@@ -54,9 +54,7 @@ if(NOT DEFINED ENABLE_LTO)
   set(ENABLE_LTO OFF)
 endif()
 
-if(NOT DEFINED DISABLE_ES2015)
-  set(FEATURE_PROFILE "es2015-subset")
-endif()
```
yorkie: The above lines should not be removed, I think adding the default FEATURE_PROFILE for ShadowNode would be better :)

profiles/jerry.profile
```diff
@@ -0,0 +1,2 @@
+CONFIG_DISABLE_ES2015_PROMISE_BUILTIN
```
yorkie: How about putting the profile on the root directory? Actually we won't own other `*.profile` files anyway.

algebrait: Will ShadowNode use the same profile file on every target later? If not, I would prefer put it on profilers directory.

yorkie: So how about profilers/default.profile?

algebrait: Directory profiles already has default.profile and minimal.profile. They are for ShadowNode modules. Perhaps default.jerry.profile is better to show it is profile for jerry.

yorkie: Why not `./profiles/jerry/default.profile`, an implicit separator would make confusion a bit to me :(

algebrait: Well done, that's it!

algebrait:
There is still a jerry unit test failed - test-snapshot.c.
Because #390 will change snapshot format, this PR will not give a temp patch.

PR #385, opened 2018-10-21
promise: optimize the performance on .then and static resolve

yorkie:
The optimized `Promise` comparison:

| versions | 1 | 2 | 3 | 4 | 5 | Avg |
|----------|--|--|--|--|--|--|
| `before#then` | 2,500 | 2,189 | 2,560 | 2,716 | 2,191 | 2,431 |
| `after#then` | 3,056 | 3,149 | 3,383 | 3,270 | 3,694 | 3,310 |
| `before#resolve` | 1,717 | 1,589 | 1,744 | 1,684 | 1,770 | 1,700 |
| `after#resolve` | 2,848 | 2,764 | 3,433 | 2,622 | 3,023 | 2,938 |

The performance slots are:

- `then`, removes to create duplicated empty function, and creates at start.
- `Promise.resolve`:
  - skips `doResolve`, and immediately calls `resolve(self, value)` instead.
  - introduces freeze promise objects, currently for `undefined`, `true` and `false`.

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/benchmark/promise/resolve-empty-array.js
+++ b/benchmark/promise/resolve-empty-array.js
@@ -1,0 +1,21 @@
+'use strict';
+
+var common = require('../common.js');
+var bench = common.createBenchmark(main, {
+  n: [128],
+});
+
+function main(opts) {
+  var n = opts.n;
+  var future = Promise.resolve([]);
+  bench.start();
+
+  for (var i = 0; i < n; ++i) {
+    future = future.then(() => {
+      return Promise.resolve([]);
+    });
+  }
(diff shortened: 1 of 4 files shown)


legendecas:
`Promise.resolve(null)` shall be included in the freed list too.

legendecas:
Shall frozen promise be `Object.freeze`-ed to prevent any possible modification on the global frozen object?

yorkie:
The CI seems pass, but GitHub doesn't get correct state.

PR #392, opened 2018-10-25
iotjs: cache parsed dump table for error.stack at runtime

yorkie:
This optimizes the performance when printing stacktrace from an error.
We did parse and generate the table every time when we want to print
the stack.

##### Checklist

- [ ] `npm test` passes

### DIFF
--- a/src/js/iotjs.js
+++ b/src/js/iotjs.js
@@ -49,49 +49,65 @@
   var module = Module.require('module');
   var fs = Module.require('fs');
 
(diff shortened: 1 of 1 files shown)


yorkie:
iotjs benchmark/run.js error

error/stack.js
error/stack.js n=128: 249.5510717566385

./build/x86_64-darwin/debug/bin/iotjs benchmark/run.js error

error/stack.js
error/stack.js n=128: 1,055.6660197330866

PR #399, opened 2018-10-30
n-api: code is optional on create error

legendecas:
- [x] `npm test` passes

According to N-API docs, second argument of napi_create_error/napi_create_type_error/napi_create_range_error shall be an optional string typed JS value.

### DIFF
--- a/src/napi/node_api_value.c
+++ b/src/napi/node_api_value.c
@@ -89,7 +89,6 @@ napi_status napi_create_object(napi_env env, napi_value* result) {
     jerry_value_t jval_code = AS_JERRY_VALUE(code);                           \
     jerry_value_t jval_msg = AS_JERRY_VALUE(msg);                             \
                                                                               \
-    NAPI_TRY_TYPE(string, jval_code);                                         \
     NAPI_TRY_TYPE(string, jval_msg);                                          \
                                                                               \
     jerry_size_t msg_size = jerry_get_utf8_string_size(jval_msg);             \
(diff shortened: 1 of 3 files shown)


src/napi/node_api_value.c
```diff
@@ -107,8 +106,11 @@ napi_status napi_create_object(napi_env env, napi_value* result) {
      * error flag.                                                            \
      */                                                                       \
     jerryx_create_handle(jval_error);                                         \
-                                                                              \
-    iotjs_jval_set_property_jval(jval_error, "code", jval_code);              \
+    /** code has to be an JS string type, thus it can not be an number 0 */   \
+    if (code != NULL) {                                                       \
+      NAPI_TRY_TYPE(string, jval_code);                                       \
+      iotjs_jval_set_property_jval(jval_error, "code", jval_code);            \
```
yorkie: Add code as MAGIC string?

src/napi/node_api_value.c
```diff
@@ -107,8 +106,11 @@ napi_status napi_create_object(napi_env env, napi_value* result) {
      * error flag.                                                            \
      */                                                                       \
     jerryx_create_handle(jval_error);                                         \
-                                                                              \
-    iotjs_jval_set_property_jval(jval_error, "code", jval_code);              \
+    /** code has to be an JS string type, thus it can not be an number 0 */   \
+    if (code != NULL) {                                                       \
```
yorkie: Validate it in JerryScript API?

legendecas: it has been validated on next line only if code is not NULL - that is code is given a value.

PR #400, opened 2018-10-30
MQTT: update event trigger timing

lolBig:
- change the trigger timing of the connect event
- add close event when the connection is disconnected, see [mqtt.js](https://github.com/mqttjs/MQTT.js#event-close)

### DIFF
--- a/test/run_pass/test_mqtt_client.js
+++ b/test/run_pass/test_mqtt_client.js
@@ -3,17 +3,22 @@
 var testHost = 'mqtt://test.mosquitto.org:1883';
 var mqtt = require('mqtt');
 var assert = require('assert');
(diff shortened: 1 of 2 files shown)


test/run_pass/test_mqtt_client.js
```diff
@@ -3,17 +3,22 @@
...
+  // trigger the close event
   client.disconnect();
   assert.equal(client._keepAliveTimer, null);
   assert.equal(client._keepAliveTimeout, null);
-  disconnected = true;
-});
-client.once('offline', function() {
-  assert.ok(disconnected);
 });
+client.once('error', function(err) {
```
legendecas: If an assertion shall be made here?

PR #401, opened 2018-10-30
process: refactor nextTick

lolBig:
- [x]   npm test passes
- [x]   tests and/or benchmarks are included

- separate the implementation of nextTick into another file
- use list instead of array to storage the callbacks
- optimizes the invocation performance of the callbacks
- handle all callbacks in a tick. in the previous implementation, a clone nextTickQueue is used to iterating, this will lead to a problem that a nextTick callback added by a calling nextTick callback can not be called in a same tick.

### DIFF
--- a/src/js/next_tick.js
+++ b/src/js/next_tick.js
@@ -1,0 +1,84 @@
+'use strict';
+
+function TickObject(callback, args) {
+  this.callback = callback;
+  this.args = args;
+}
+
+function NextTickQueue() {
+  this.head = null;
+  this.tail = null;
+  this.length = 0;
+}
+
+NextTickQueue.prototype.push = function push(v) {
+  var entry = { data: v, next: null };
+  if (this.length > 0)
+    this.tail.next = entry;
+  else
+    this.head = entry;
+  this.tail = entry;
+  ++this.length;
+};
+
+NextTickQueue.prototype.shift = function shift() {
+  if (this.length === 0)
+    return;
+  var ret = this.head.data;
+  if (this.length === 1)
+    this.head = this.tail = null;
+  else
+    this.head = this.head.next;
+  --this.length;
+  return ret;
+};
+
+NextTickQueue.prototype.clear = function clear() {
+  this.head = null;
+  this.tail = null;
+  this.length = 0;
+};
+
+var nextTickQueue = new NextTickQueue();
+
+module.exports.nextTick = function nextTick(callback) {
+  var args;
+  switch (arguments.length) {
+    case 1: break;
+    case 2: args = [arguments[1]]; break;
+    case 3: args = [arguments[1], arguments[2]]; break;
+    case 4: args = [arguments[1], arguments[2], arguments[3]]; break;
+    default:
+      args = new Array(arguments.length - 1);
+      for (var i = 1; i < arguments.length; i++)
+        args[i - 1] = arguments[i];
+  }
+  var tickObject = new TickObject(callback, args);
+  nextTickQueue.push(tickObject);
+};
+
+module.exports._onNextTick = function _onNextTick() {
+  while (nextTickQueue.length > 0) {
+    var tickObject = nextTickQueue.shift();
+    var callback = tickObject.callback;
+    var args = tickObject.args;
+    if (args === undefined) {
+      callback();
+    } else {
+      switch (args.length) {
+        case 1:
+          callback(args[0]);
+          break;
+        case 2:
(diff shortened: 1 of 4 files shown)


src/js/next_tick.js
```diff
@@ -0,0 +1,82 @@
...
+      switch (args.length) {
+        case 1: callback(args[0]); break;
+        case 2: callback(args[0], args[1]); break;
+        case 3: callback(args[0], args[1], args[2]); break;
+        default: callback.apply(undefined, args); break;
+      }
+    }
+  }
+  return false;
+};
```
yorkie: Next tick should not be a separate module, propose that add a internal module `process` for that.

yorkie:
BTW, I think the `setImmediate` should be included in this patch, too.

src/js/next_tick.js
legendecas: It's reasonable for a separate module since it does not depends on other `process` things at all.

As next_tick is mounted on `process` instance, `internal/process/next_tick.js` might be better.

yorkie: Yea, this is what I mean. Unfortunately I doesn't own the real internal module, which causes the below is working in user land:

```js
var nextTick = require('next_tick');
```

We should just rename the module to be `internal/process/next_tick` here?

src/modules.json
```diff
@@ -293,6 +294,9 @@
       "js_file": "js/net.js",
       "require": ["assert", "events", "stream", "tcp", "util"]
     },
+    "next_tick": {
```
yorkie: As I have pointed at https://github.com/Rokid/ShadowNode/pull/401/files#r229584228, this needs to be changed to `internal/process/next_tick`.

src/js/next_tick.js
lolBig: emmmmm, should be like this

benchmark/process/nextTick_parallel.js
```diff
@@ -0,0 +1,19 @@
...
+var common = require('../common.js');
+var bench = common.createBenchmark(main, {
+  n: [1024],
+});
+
+function main(opts) {
+  var n = opts.n;
+  process.nextTick(function() {
+    bench.start();
+  })
```
yorkie:
```suggestion
process.nextTick(function() {
    bench.start();
    for (var i = 0; i < n; ++i) {
      process.nextTick(function(i) {}, i);
    }
    process.nextTick(function() {
      bench.end(n);
    });
  });
});
```

src/js/internal/process/next_tick.js
```diff
@@ -0,0 +1,43 @@
...
+
+module.exports.nextTick = function nextTick(callback) {
+  var args;
+  switch (arguments.length) {
+    case 1: break;
+    case 2: args = [arguments[1]]; break;
+    case 3: args = [arguments[1], arguments[2]]; break;
+    case 4: args = [arguments[1], arguments[2], arguments[3]]; break;
+    default:
+      args = Array.prototype.slice.call(arguments, 1);
```
yorkie: We should cover the above cases in our benchmark, too.

src/js/internal/process/next_tick.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+var nextTickQueue = [];
+
+module.exports.nextTick = function nextTick(callback) {
+  var args;
+  switch (arguments.length) {
+    case 1: break;
+    case 2: args = [arguments[1]]; break;
```
yorkie: We don't need the array actually.

yorkie: With the following suggestion:

```suggestion
    case 2: args = arguments[1]; break;
```

benchmark/process/nextTick_create.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
```
yorkie: Suggestion: nextTick_create -> next_tick_create

src/modules.json
```diff
@@ -293,6 +294,9 @@
       "js_file": "js/net.js",
       "require": ["assert", "events", "stream", "tcp", "util"]
     },
+    "internal_process_next_tick": {
```
yorkie: Why not `internal/process/next_tick'?

legendecas: right now module names are stored in cmake variable, which could not include slash "/" nor back slash"\".

yorkie:
Holy shit awesome improvements on performance! Thanks @lolBig @legendecas

PR #402, opened 2018-10-31
v0.10.13 Release Proposal

yorkie:
In this release, we have landed:

- build: use CMAKE_INSTALL_PREFIX instead of custom install command.
- build: add dependencies for iotjs.cmake which depends on others.
- events: cannot read property 'newListener' of undefined.
- MQTT: update event trigger timing.
- n-api: code is optional on create error.

And we supported the clang-format-7 on TravisCI, too.


yorkie:
@legendecas don't merge/rebase via GitHub button for this type of patch, rebasing commits in PR would make the version branch only leaves those version commits.

PR #403, opened 2018-11-01
mqtt: fixed the unconnectable problem caused by setting a wrong isConnected flag

lolBig:
this flag causes the MQTT ACK request to fail


src/js/mqtt.js
```diff
@@ -47,6 +47,7 @@ function MqttClient(endpoint, options) {
     protocolVersion: 4,
     pingReqTimeout: 10 * 1000,
   }, options);
+  this._isTCPConnected = false;
```
yorkie: For naming with Node.js, `isSocketConnected` might be proper :)

PR #407, opened 2018-11-08
uv: drop code path for epoll_pwait-less kernels

yorkie:
Merges https://github.com/bnoordhuis/libuv/commit/f43c66 into tuv, that:

> Removes the support for kernels that don't support epoll_pwait()
> now that the minimum requirements are kernel 2.6.32 + glibc 2.12.

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/deps/libtuv/src/unix/linux-core.c
+++ b/deps/libtuv/src/unix/linux-core.c
@@ -191,8 +191,6 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
    * that being the largest value I have seen in the wild (and only once.)
    */
   static const int max_safe_timeout = 1789569;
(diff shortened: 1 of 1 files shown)


yorkie:
This https://github.com/libuv/libuv/commit/9208df0c93c270cf3eb48a910e50b70c9e48b150 might be needed to be landed on tuv.

yorkie:
This actually attempts to address a crash on `uv__io_poll` which possibly dues to the `pwait` state that has addressed at https://github.com/libuv/libuv/pull/1372, and have integration testing over 72 hours.

legendecas:
AFAIK, this patch only addresses a few issues on mips build?

PR #408, opened 2018-11-08
dbus: add an external param `waitOnDelay` to handle service unavailable

yorkie:
Sometimes the D-Bus service is unavailable at bootstraping the os, so the client will get a connection with NULL, and the following operations with this connection would make failures even abortions.

This patch intros a synchronous wait, that the user could gives a max timeout, within that, ShadowNode attempts to connect and check the following errors:

- `org.freedesktop.DBus.Error.FileNotFound`
- `org.freedesktop.DBus.Error.Failed`

If one of them, client would sleep 100ms and retry until the current time is over the given max timeout. Otherwise it breaks the loop for other errors like "Address does not contain a colon".

##### Checklist

- [x] `npm test` passes


src/modules/iotjs_module_dbus.c
```diff
@@ -423,22 +422,48 @@ JS_FUNCTION(GetBus) {
...
+    if (_this->connection != NULL)
+      break;
+
+    fprintf(stderr, "DBUS(%s) %s\n", error.name, error.message);
+    if (wait <= 0 ||
+        (strcmp(error.name, "org.freedesktop.DBus.Error.FileNotFound") != 0 &&
+         strcmp(error.name, "org.freedesktop.DBus.Error.Failed") != 0)) {
+      break;
+    } else {
+      usleep(IOTJS_DBUS_CONN_DELAY);
```
legendecas: Is this `usleep` sleeping on main thread?

yorkie: See https://github.com/Rokid/ShadowNode/pull/408/files#diff-ddc1ef60d9b05e87f82ea405bdf4e9aaR433, yes, it blocks the main thread the `max_timeout` ms in bad case.

yorkie:
> Also unit used in usleep is microsecond rather than millisecond, yet it has been used with wait in millisecond.

@legendecas See this line: https://github.com/Rokid/ShadowNode/pull/408/files#diff-ddc1ef60d9b05e87f82ea405bdf4e9aaR27, I have converted it to millisecond, do you have function suggestion that accepts the value in millisecond?

src/modules/iotjs_module_dbus.c
legendecas: If it is more suitable that throwing an error to JS and use setTimeout which is based on uv timers? or directly register an uv timer in C land.

lolBig: agreed with @legendecas

legendecas: Breaking API is more acceptable to me to sleep in a library. Is there any other strong objections against throwing an error or provide an async API?

yorkie: > Breaking API is more acceptable to me than sleeping in a library

We should discuss on a context, non-blocking API might be better than current implementation is that I was agreeing, too. That's why I commented a FIXME on my source code. The context is that introducing a breaking change causes that could not be landed at v0.10.x.

src/js/dbus.js
```diff
@@ -512,13 +513,14 @@ ServiceInterface.prototype.update = util.deprecate(function() {
...
  * @method getBus
  * @param {String} name
+ * @param {Number} delayOnWait
  */
-function getBus(name) {
+function getBus(name, delayOnWait) {
   if (_busInstance === false) {
     throw new Error('dbus connection has been destroyed');
   }
   if (_busInstance === null) {
```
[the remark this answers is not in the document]
yorkie: May I address in another PR? This is not relevant on the source changed.

src/modules/iotjs_module_dbus.c
legendecas: If a new async api shall not be introduced in a patch version, an error shall be thrown rather than blocking implicitly which would definitely introduce unpredictable behaviors.

legendecas: Error handling in establishing a connection might not be an edge case as the comment saying IMO.

yorkie:
Let me re-state again, Node.js owns such many blocking APIs like `fs.openSync`, `fs.closeSync` and more in core modules.

@legendecas @lolBig If you were considering the blocking-ones, does document this `getBus(type, waitOnDelay)` looks good?

yorkie:
After having discussion on boarding, we decided to close this PR, and land this on v0.10.11 only as a hotfix.

PR #409, opened 2018-11-08
uv, os: implement os.{get,set}Priority functions

yorkie:
This implements the following APIs:

- https://github.com/nodejs/node/blob/master/doc/api/os.md#osgetprioritypid
- https://github.com/nodejs/node/blob/master/doc/api/os.md#ossetprioritypid

With the group of APIs, we are able to tweak the priority level on demand at runtime.

##### Checklist

- [x] `npm test` passes


src/js/os.js
```diff
@@ -16,6 +16,41 @@ exports.freemem = function() {
   return native.getFreeMem();
 };
 
+exports.getPriority = function(pid) {
+  if (pid === undefined)
```
lolBig:
```suggestion
  if (pid === undefined) {
```

lolBig: missing `{`

src/js/os.js
```diff
@@ -16,6 +16,41 @@ exports.freemem = function() {
...
+    priority = pid;
+    pid = 0;
+  } else {
+    pid = Number(pid);
+  }
+
+  if (typeof priority !== 'number') {
+    throw new TypeError('priority must be a number');
+  }
+  if (priority > 19) {
```
lolBig: use `UV_PRIORITY_LOW` and `UV_PRIORITY_HIGHST` defined  from  UV instead of literal ?

yorkie: Actually those defines are for using at Windows, I still leave those on uv.h is to acquire consistence.

lolBig: seems those defines are using at Linux too, see setPriority [doc](https://linux.die.net/man/2/setpriority)

yorkie: Could you please reference the code/lines in details? Can not search any that defines from your link :(

yorkie:
The CI seems green now, @lolBig @legendecas please see again :)

src/js/os.js
```diff
@@ -16,6 +16,42 @@ exports.freemem = function() {
...
+    priority = 19;
+  } else if (priority < -20) {
+    priority = -20;
```
legendecas: Should an error been thrown instead of silently rounded?

src/js/os.js
```diff
@@ -16,6 +18,42 @@ exports.freemem = function() {
...
+    pid = Number(pid);
+  }
+
+  if (typeof priority !== 'number') {
+    throw new TypeError('priority must be a number');
+  }
+  if (priority > constants.UV_PRIORITY_LOW ||
+    priority < constants.UV_PRIORITY_HIGHEST) {
+    throw new RangeError(
+      'The value of "priority" is out of range. It must be > -20 && < 19.');
```
lolBig: use constants instead of -20 or 19

PR #406, opened 2018-11-08
process: invokes setImmediate on uv check phase

legendecas:
- [x] `npm test` passes
- [x] tests and/or benchmarks are included

Also enables uv check/prepare handles in libtuv.

PR #411, opened 2018-11-09
n-api: thread safe functions

legendecas:
- [x] `npm test` passes

Unit tests depends on #406.

Introduce thread safe functions into N-API. Simplify complex native add-ons to notify JavaScripts in a work thread without blocking work thread nor main thread.

### DIFF
--- a/src/internal/node_api_internal.h
+++ b/src/internal/node_api_internal.h
@@ -19,6 +19,7 @@
 #include "iotjs_def.h"
 #include "jerryscript-ext/handle-scope.h"
 #include "jerryscript.h"
+#define NAPI_EXPERIMENTAL
 #include "internal/node_api_internal_types.h"
 #include "node_api.h"
 
--- a/src/internal/node_api_internal_types.h
+++ b/src/internal/node_api_internal_types.h
@@ -134,9 +134,9 @@ struct iotjs_threadsafe_function_s {
   napi_threadsafe_function_call_js call_js_cb;
 
   napi_async_context async_context;
-  uv_async_t* async_handle;
-  uv_cond_t* async_cond;
-  uv_mutex_t* op_mutex;
+  uv_async_t async_handle;
+  uv_cond_t async_cond;
+  uv_mutex_t op_mutex;
 
   iotjs_tsfn_invocation_t* invocation_head;
   iotjs_tsfn_invocation_t* invocation_tail;
--- a/src/internal/node_api_internal_types.h
+++ b/src/internal/node_api_internal_types.h
@@ -34,6 +34,8 @@ typedef struct iotjs_function_info_s iotjs_function_info_t;
 typedef struct iotjs_napi_env_s iotjs_napi_env_t;
 typedef struct iotjs_object_info_s iotjs_object_info_t;
 typedef struct iotjs_reference_s iotjs_reference_t;
+typedef struct iotjs_tsfn_invocation_s iotjs_tsfn_invocation_t;
+typedef struct iotjs_threadsafe_function_s iotjs_threadsafe_function_t;
 
 typedef enum {
   napi_module_load_ok = 0,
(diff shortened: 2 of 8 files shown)


yorkie:
Let's remove the ugly prefix on title and start using the label `work-in-progress` :)

src/internal/node_api_internal.h
```diff
@@ -19,6 +19,7 @@
 #include "iotjs_def.h"
 #include "jerryscript-ext/handle-scope.h"
 #include "jerryscript.h"
+#define NAPI_EXPERIMENTAL
```
yorkie: Where does this be used?

src/internal/node_api_internal_types.h
```diff
@@ -114,4 +116,33 @@ struct iotjs_async_context_s {
...
+
+  void* thread_finalize_data;
+  napi_finalize thread_finalize_cb;
+  void* context;
+  napi_threadsafe_function_call_js call_js_cb;
+
+  napi_async_context async_context;
+  uv_async_t* async_handle;
+  uv_cond_t* async_cond;
+  uv_mutex_t* op_mutex;
```
yorkie: I think we could use `uv_{async/cond/mutex}_t` instead of pointers to reduce the times of malloc?

src/napi/node_api_tsfn.c
```diff
@@ -0,0 +1,314 @@
...
+
+  bool napi_async_inited = false, uv_async_inited = false,
+       uv_cond_inited = false, uv_mutex_inited = false, func_acquired = false;
+
+  napi_async_context async_context;
+  NAPI_INTERNAL_CALL(napi_async_init(env, async_resource, async_resource_name,
+                                     &async_context));
+  napi_async_inited = true;
+
+  uv_async_t* async_handle = IOTJS_ALLOC(uv_async_t);
```
yorkie: This could be removed if we directly use struct instead of its pointer :)

src/napi/node_api_tsfn.c
```diff
@@ -0,0 +1,314 @@
...
+  iotjs_environment_t* iotjs_env = iotjs_environment_get();
+  uv_loop_t* iotjs_loop = iotjs_environment_loop(iotjs_env);
+  int uv_status = uv_async_init(iotjs_loop, async_handle, tsfn_async_callback);
+  if (uv_status != 0) {
+    goto clean;
+  }
+  uv_async_inited = true;
+
+  uv_cond_t* async_cond = IOTJS_ALLOC(uv_cond_t);
```
yorkie: Ditto.

src/napi/node_api_tsfn.c
```diff
@@ -0,0 +1,314 @@
...
+  uv_status = uv_cond_init(async_cond);
+  if (uv_status != 0) {
+    goto clean;
+  }
+  uv_cond_inited = true;
+
+
+  uv_mutex_t* op_mutex = IOTJS_ALLOC(uv_mutex_t);
```
yorkie: Ditto.

src/napi/node_api_tsfn.c
```diff
@@ -0,0 +1,314 @@
...
+    uv_cond_destroy(async_cond);
+  }
+  if (uv_mutex_inited) {
+    uv_mutex_destroy(op_mutex);
+  }
+  NAPI_RETURN_WITH_MSG(napi_generic_failure,
+                       "Unexpected error on napi_create_threadsafe_function.");
+}
+
+napi_status napi_get_threadsafe_function_context(napi_threadsafe_function func,
```
yorkie: This function could be inlined?

src/napi/node_api_tsfn.c
```diff
@@ -0,0 +1,314 @@
...
+  uv_mutex_lock(tsfn->op_mutex);
+  napi_status ret_status = napi_ok;
+
+
+  if (tsfn->max_queue_size != 0 && tsfn->queue_size == tsfn->max_queue_size) {
+    switch (is_blocking) {
+      case napi_tsfn_nonblocking:
+        ret_status = napi_queue_full;
+        goto clean;
+        break;
```
yorkie: The `break` statement should be unreachable :(

test/napi/napi_tsfn.test.js
```diff
@@ -0,0 +1,208 @@
+'use strict';
```
yorkie: Are this file a copy from https://github.com/Rokid/ShadowNode/blob/master/test/addons-napi/test_threadsafe_function/test.js? If yes, suggest that open an issue on [nodejs/napi-test-suite](https://github.com/nodejs/napi-test-suite) to track how should we address this with N-API team.

src/internal/node_api_internal.h
legendecas: It's used in `node_api.h`, in which declares napi experimental functions if `NAPI_EXPERIMENTAL` has been defined.

src/napi/node_api_tsfn.c
legendecas: This function is declared in `node_api.h`. We should not change its function signature.

yorkie: I see.

PR #412, opened 2018-11-09
n-api: napi_create_symbol

legendecas:
- [x] `npm test` passes

https://github.com/nodejs/node-addon-api/blob/master/napi-inl.h#L72

node-addon-api used symbol in most of places. To support node-addon-api, napi_create_symbol has to be polyfilled.


yorkie:
SGTM if there is more test suite could be included :)

legendecas:
Still, symbol has to be supported in JerryScript, since unique object could not be a valid key and by default it is converted to a string if it's used as a key.

PR #415, opened 2018-11-19
util: IOTJS_ASSERT prints stack trace on macOS

legendecas:
- [x] `npm test` passes

Currently stack trace only prints on linux platform. While macOS actually support backtrace too, this PR enables backtrace on macOS platform.

### DIFF
--- a/src/iotjs_util.c
+++ b/src/iotjs_util.c
@@ -103,6 +103,7 @@ void print_stacktrace() {
   for (int idx = 0; idx < size; ++idx) {
     fprintf(stderr, "%s\n", bt_strs[idx]);
   }
+  free(bt_strs);
 #endif // (defined(__linux__) || defined(__APPLE__))
 }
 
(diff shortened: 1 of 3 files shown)


src/iotjs_util.c
```diff
@@ -95,39 +95,15 @@ void iotjs_buffer_release(char* buffer) {
...
-  char command[256];
-
-  int nptrs = backtrace(buffer, numOfStackTrace);
-  char** strings = backtrace_symbols(buffer, nptrs);
-
-  if (strings == NULL) {
-    perror("backtrace_symbols");
-    exit(EXIT_FAILURE);
+#if (defined(__linux__) || defined(__APPLE__))
+  void* bt[IOTJS_BACKTRACE_LEN];
```
yorkie: @legendecas May I ask you to revert the name to buffer?

legendecas: its not a buffer. this is a storage of backtrace.

yorkie: So what's the `bt` for?

PR #416, opened 2018-11-20
n-api: update n-api headers/test sets to LTS version 10.13.0

legendecas:
- [x] `npm test` passes


include/node_api.h
```diff
@@ -697,6 +697,12 @@ NAPI_EXTERN napi_status napi_get_value_bigint_words(napi_env env,
                                                     int* sign_bit,
                                                     size_t* word_count,
                                                     uint64_t* words);
+NAPI_EXTERN napi_status napi_add_finalizer(napi_env env,
```
yorkie: A new API? We should document this function at our unsupported list if this is not implemented :)

PR #417, opened 2018-11-20
process: set immediate shall start an idle handle to activate uv_loop

legendecas:
- [x] `npm test` passes


yorkie:
Good nice, LGTM if CI is green 👍

yorkie:
The CI seems fail(timeout) at `TIMEOUT: napi_reference.test.js`.

legendecas:
@yorkie setImmediate has been refactored so that uv loop could not be blocked on polling if there is only check handles.

yorkie:
Because this depends on #406, labeled as `minor` as well.

Issue #414, opened 2018-11-17
os: networkInterfaces returns bcast address


yorkie:
See http://man7.org/linux/man-pages/man3/getifaddrs.3.html, we need to expose both `ifu_broadaddr` and `ifu_dstaddr` at some use case.

Txiaozhe:
It means we need add this two function into ShadowNode? I have read the web page but not very distinct yet.
In this web:
```c
int getifaddrs(struct ifaddrs **ifap);
void freeifaddrs(struct ifaddrs *ifa);
```
it means I need to use this two function to achieve `ifu_broadaddr` and `ifu_dstaddr` functions?

yorkie:
Nope, we should return those from the same function.

Txiaozhe:
You mean, we will use `ifu_broadaddr` and `ifu_dstaddr` to return same result as  
```c
int getifaddrs(struct ifaddrs **ifap);
void freeifaddrs(struct ifaddrs *ifa);
```
?

yorkie:
Basically yes, it looks like the following:

```shell
$ node -e "os.networkInterfaces()"
> { "wlan0": 
    [ { address: '127.0.0.1',
         network: '255.0.0.0',
         family: 'IPv4',
         mac: '00:00:00:00:00:00',
         broadcast: '127.0.0.2' } ]
}
```

Txiaozhe:
Yes, I will try to achieve it.

Txiaozhe:
Does this two methods:
```c
int getifaddrs(struct ifaddrs **ifap);
void freeifaddrs(struct ifaddrs *ifa);
```
 only support Linux platform? Can I find other ways to support it in mac.

yorkie:
They should be supported on darwin, too.

Txiaozhe:
Yes, I confirmed it. Thanks!

Txiaozhe:
```c
#include <stdio.h>
#include <ifaddrs.h>
#include <arpa/inet.h>
#include <netdb.h>
#include <stdlib.h>
#include <stdbool.h>

int getNameInfo(struct sockaddr *addr, int size_of_saddr_in, char *host) {
    return getnameinfo(addr, size_of_saddr_in, host, NI_MAXHOST, NULL, 0, NI_NUMERICHOST);
}

int getSubnetMask()
{
    struct sockaddr_in *sin = NULL;
    struct sockaddr_in6 *sin6 = NULL;
[...]

PR #420, opened 2018-11-23
os: bcast address interface

Txiaozhe:
##### Checklist

- [x] `npm test` passes


Txiaozhe:
#414

deps/libtuv/src/unix/darwin.c
```diff
@@ -293,6 +293,14 @@ int uv_interface_addresses(uv_interface_address_t** addresses, int* count) {
       address->netmask.netmask4 = *((struct sockaddr_in*) ent->ifa_netmask);
     }
 
+    if (ent->ifa_dstaddr) {
+      if (ent->ifa_dstaddr->sa_family == AF_INET6) {
+        address->broadcast.broadcast6 = *((struct sockaddr_in6*) ent->ifa_dstaddr);
+      } else {
+        address->broadcast.broadcast4 = *((struct sockaddr_in*) ent->ifa_dstaddr);
+      }
+    }
```
yorkie: You should change the other platforms, too.

test/run_pass/test_os_network_interfaces.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+var os = require('os');
+
+console.log(os.networkInterfaces());
```
yorkie: You should assert the fields, just like what the Node.js does at https://github.com/nodejs/node/blob/master/test/parallel/test-os.js#L131-L172.

yorkie:
Label this as `push upstream`, so when this gets done, I suggest that @Txiaozhe you could push this change on libuv and Node.js core :)

test/run_pass/test_os_network_interfaces.js
Txiaozhe: Yes, I will

test/run_pass/test_os_network_interfaces.js
```diff
@@ -0,0 +1,21 @@
...
+
+var assert = require('assert');
+var os = require('os');
+
+var family = {
+  IPv4: 1,
+  IPv6: 1
+};
+
+var net_info = os.networkInterfaces();
```
yorkie: Ensure the interfaces array's length is >0 :)

test/run_pass/test_os_network_interfaces.js
```diff
@@ -0,0 +1,21 @@
...
+
+for (var net_name in net_info) {
+  net_info[net_name].forEach(net => {
```
yorkie: net_name -> name

yorkie:
By the way, you should add your test case to https://github.com/Rokid/ShadowNode/blob/master/test/testsets.json in alphabeta.

Txiaozhe:
Thanks, has been fixed.

test/run_pass/test_os_network_interfaces.js
```diff
@@ -0,0 +1,22 @@
...
+for (var name in net_info) {
+  assert(net_info[name].length);
+  net_info[name].forEach(net => {
+    assert.equal(typeof net.address, 'string');
+    assert.equal(typeof net.netmask, 'string');
+    assert.equal(family[net.family], 1);
+    assert.equal(typeof net.broadcast, 'string');
+    assert.equal(typeof net.mac, 'string');
+  });
+}
```
yorkie: Could you please just move the Node.js [test](https://github.com/nodejs/node/blob/master/test/parallel/test-os.js#L131-L172) into this repository?

Txiaozhe: Yes, I will try it.

test/common/index.js
```diff
@@ -61,6 +89,18 @@ function expectsError(fn, exact) {
...
+  isLinuxPPCBE: isLinuxPPCBE,
+  isSunOS: isSunOS,
+  isFreeBSD: isFreeBSD,
+  isOpenBSD: isOpenBSD,
+  isLinux: isLinux,
+  isOSX: isOSX,
+  isOSXMojave: isOSXMojave,
+
+  is: is,
+
```
yorkie: @Txiaozhe No need to add these lines.

Txiaozhe: Awkward! I will do it.

yorkie: Never mind, you are always welcome :)

test/run_pass/test_os.js
```diff
@@ -0,0 +1,47 @@
...
+var platform = os.platform();
+common.is.string(platform);
+assert.ok(platform.length > 0);
+
+var EOL = os.EOL;
+if (common.isWindows) {
+  assert.strictEqual(EOL, '\r\n');
+} else {
+  assert.strictEqual(EOL, '\n');
+}
```
yorkie: We do not need the above tests at this PR, could you please remove them?

test/run_pass/test_os.js
```diff
@@ -0,0 +1,47 @@
...
+    
+    var expected = [{
+      address: '127.0.0.1',
+      netmask: '255.0.0.0',
+      family: 'IPv4',
+      broadcast: "127.0.0.1",
+      mac: '00:00:00:00:00:00'
+    }];
+
+    assert.deepStrictEqual(actual, expected);
```
yorkie: Unix/Linux also needs to cover :)

src/modules.json
```diff
@@ -56,7 +56,11 @@
     },
     "assert": {
       "js_file": "js/assert.js",
-      "require": ["util"]
+      "require": ["util", "comparisons"]
+    },
+    "comparisons": {
```
legendecas: An internal module doesn't been expected to be in public scope. An internal prefix might help.

Txiaozhe: Yes, I change it to internal_comparisons. Thanks! :)

test/run_pass/test_os.js
Txiaozhe: I will do!

test/run_pass/test_os.js
Txiaozhe: You mean I should comparisons Unix/Linux such as:
```c
if (Unix) {
    ....
} else if (Linux) {
    ....
}
```
?

yorkie: Nope, you are only testing this function on MacOS.

Txiaozhe: Awkward, I found my fault~

src/internal/util/comparisons.js
```diff
@@ -0,0 +1,98 @@
...
+  if (val1 && val1 === val2) {
+    return true;
+  }
+
+  return strictDeepEqual(val1, val2);
+}
+
+module.exports = {
+  isDeepStrictEqual: isDeepStrictEqual
+};
```
yorkie: Also, no need to add this file, please revert it, and make every pull request do simply.

Txiaozhe:
> Label this as `push upstream`, so when this gets done, I suggest that @Txiaozhe you could push this change on libuv and Node.js core :)

This feature was discussed in [node issues: 23437](https://github.com/nodejs/node/issues/23437), but it seems to have been rejected.

yorkie:
@Txiaozhe Nope, it's still open for changes, I think that we could continue to work on @pprindeville's patch with the comment from @addaleax :)

Txiaozhe:
Yes, I will try!

yorkie:
Also, please clean up your branch firstly, we won't merge patch which commits > 10 :)

Txiaozhe:
> Also, please clean up your branch firstly, we won't merge patch which commits > 10 :)

Yes, I will.

PR #425, opened 2018-11-27
jerry: pass and enable jerry-test-suite

algebrait:
##### Checklist

- [x] `npm test` passes
- [x] tests and/or benchmarks are included
    jerry-test-suite pass now
- [ ] documentation is changed or added

This PR will fix #234

PR #429, opened 2018-11-29
test: fix wrong travis diff target introduced by #425

algebrait:
##### Checklist

- [x] `npm test` passes


legendecas:
Found out that reversed base version works fantastically: `git diff --name-only master...HEAD`

PR #442, opened 2018-12-12
add local server for protocol test

lolBig:
add local server for protocol test, avoid failure for external echo server reasons

supported the following protocol server:
- websocket(with ssl)
- mqtt(with ssl)

checklist
- [x] `npm test` passes
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added

PR #453, opened 2018-12-22
mqtt: support `will` options for sending messages while being disconnected

yorkie:
The MQTT.js supports the `will` option to send messages to a given topic, thus peers could get the unexpected disconnection.

Because the MQTT documentation did reference the source(MQTT.js), then no docs is needed to be updated. And I will cherry-pick on v0.11.x as hot-fix when this gets merged.

##### Checklist

- [x] `npm test` passes


yorkie:
This depends on #442, so I labeled that as "land-on-v0.10.x", either :)

src/modules/iotjs_module_mqtt.c
```diff
@@ -40,46 +40,62 @@ JS_FUNCTION(MqttConstructor) {
...
+    MQTTPacket_willOptions willOpts = MQTTPacket_willOptions_initializer;
+    jerry_value_t topicName =
+        iotjs_jval_get_property(will, IOTJS_MAGIC_STRING_TOPIC);
+    jerry_value_t message =
+        iotjs_jval_get_property(will, IOTJS_MAGIC_STRING_PAYLOAD);
+    jerry_value_t retained =
+        iotjs_jval_get_property(will, IOTJS_MAGIC_STRING_RETAIN);
+    jerry_value_t qos = iotjs_jval_get_property(will, IOTJS_MAGIC_STRING_QOS);
+    willOpts.retained = iotjs_jval_as_boolean(retained);
+    willOpts.qos = iotjs_jval_as_number(qos);
```
lolBig: missing release after iotjs_jval_get_property

src/modules/iotjs_module_mqtt.c
```diff
@@ -40,46 +40,62 @@ JS_FUNCTION(MqttConstructor) {
...
-    mqttstr.lenstring.data = strdup((char*)iotjs_string_data(&str));
-    mqttstr.lenstring.len = (int)iotjs_string_size(&str);
-    options.clientID = mqttstr;
-    iotjs_string_destroy(&str);
+#define MQTT_OPTION_ASSIGN_FROM(owner, name)                           \
+  do {                                                                 \
+    if (jerry_value_is_string(name)) {                                 \
+      iotjs_string_t str = iotjs_jval_as_string(name);                 \
+      MQTTString mqttstr = MQTTString_initializer;                     \
+      mqttstr.lenstring.data = strdup((char*)iotjs_string_data(&str)); \
```
lolBig: data from `strdup` should be freed

lolBig: since strdup is terminated by '\0', binary data like payload cannot be fully copied, the message sent will be incomplete, copy via strndup is recommended

yorkie: May I ask you a github suggestion, please?

PR #458, opened 2018-12-24
 module: support require.cache

ErosZy:
##### Checklist

- [ ] `npm test` passes


src/js/module.js
```diff
@@ -211,19 +215,19 @@ iotjs_module_t.normalizePath = function(path) {
...
+  var modPath = iotjs_module_t.resolveModPath(id, parent);
   var cachedModule = iotjs_module_t.cache[modPath];
   if (cachedModule) {
     iotjs_module_t.curr = modPath;
     return cachedModule.exports;
   }
 
+  if (process.builtin_modules[id]) {
+    iotjs_module_t.curr = id;
+    return Native.require(id);
```
yorkie: It seems nobody write the builtin module into cache.

ErosZy: yeah, i review my code, it don't seems necessary. when rewrite require.cache with native module id, it returns in the code logic above immediately. I see Native.require have already cache, so pass it, never mind :)

Issue #448, opened 2018-12-19
module: require.main is not identical to main module


legendecas:
* **Version**: 0.11.x
* **Platform**: all
* **Subsystem**: module

Run the following script with iotjs directly:
```js
console.log(require.main === module) // false
```

Txiaozhe:
test/hhhh.js
```js
'use strict';

console.log(require.main);
console.log(module);
```
when I run
```shell
$ iotjs test/hhhh.js
```

print:
```json
{
  "id": ".",
  "exports": {},
  "parent": null,
  "filename": "test\/hhhh.js"
}
{
  "id": "test\/hhhh.js",
  "exports": {},
  "filename": "\/Users\/jimtang\/GitHub\/ShadowNode\/test\/hhhh.js",
  "parent": null,
  "dirs": [
    "\/Users\/jimtang\/GitHub\/ShadowNode\/test\/"
  ]
}
```

do you mean I should get two same objects for `module.main` and `module`?

legendecas:
Yes the two shall be one identical object if the script is the entry.

PR #466, opened 2018-12-27
module: make require.main identical to main module

Txiaozhe:
##### Checklist

- [x] `npm test` passes


Txiaozhe:
#448

ErosZy:
Maybe need more tests. In node v8.x doc described:

`
For a file foo.js, this will be true if run via node foo.js, but false if run by require('./foo').
`
But we only test first situation. `require('foo.js')` not tests :)

yorkie:
+1 on @ErosZy's suggestion :)

Txiaozhe:
Yes, It is rigorous, thanks @ErosZy, I will try it.

yorkie:
BTW, I think we could push this to upstream(IoT.js).

Txiaozhe:
> BTW, I think we could push this to upstream(IoT.js).

Good idea!

test/run_pass/test_require_main_foo.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
```
legendecas: This test is not included in the test sets :(. 

As a suggestion, the file could be put in fixtures folder and then being required to cover the false case.

Txiaozhe: This only be invoked by `test_require_main.js`, but I have not found the `fixtures` folder, will I create one?

legendecas: Samples could be found: `test/run_pass/require1`, `test/run_pass/require2`, `test/run_pass/require3`

Txiaozhe: Yes, it's concise and clear, I will do it.

src/js/module.js
```diff
@@ -250,6 +250,10 @@ iotjs_module_t.load = function(id, parent) {
   iotjs_module_t.cache[modPath] = module;
   iotjs_module_t.curr = modPath;
 
+  if (isMain) {
+    mainModule = module;
```
yorkie: Miss https://github.com/nodejs/node/blob/master/lib/internal/modules/cjs/loader.js#L547

src/js/module.js
```diff
@@ -340,8 +344,8 @@ iotjs_module_t.runMain = function() {
     var fn = process.debuggerSourceCompile();
     fn.call();
   } else {
-    var filename = mainModule.filename = process.argv[1];
-    mainModule.exports = iotjs_module_t.load(filename, null);
+    var filename = process.argv[1];
+    iotjs_module_t.load(filename, null, true);
```
yorkie:
```js
iotjs_module_t.load(process.argv[1], null, true);
```

test/run_pass/require_main/foo.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+module.exports = function() {
+  return require.main === module;
+}
```
yorkie: This file is not testing anything, move this to fixtures directory.

Txiaozhe: But I have move it into `run_pass/require_main` dir, it is fault?

yorkie: Misunderstood, but a suggestion is the name to be require4 or add your test to exists require dir.

Txiaozhe: Maybe it is more appropriate, I will do it.

yorkie: Thanks, the Node.js team maintains a huge tests now, ShadowNode doesn't want to do like that, thus we should own a better group for our tests.

Txiaozhe: yeah, it is necessary！:)

test/run_pass/require4/foo.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+module.exports = function() {
+  return require.main === module;
+}
```
yorkie: Missing the last `;`.

Txiaozhe: Yeah, But out lint not scentout.

Issue #330, opened 2018-09-06
module: error on requiring 0 size js file


legendecas:
* **Version**: 0.10.0
* **Platform**: all
* **Subsystem**: module

if requiring a zero sized js source file, an error was raised and shadow node exited with non-zero code.

```
Error: Could not load the source.
    at anonymous (module:265:36)
    at anonymous (module:213:23)
    at anonymous (module:300:26)
```

yorkie:
Assigned to you :)

Txiaozhe:
Yes, I will take a look? Thanks:)

Txiaozhe:
In src/modules/iotjs_module_process.c: line 149
```c
if (source == NULL || size == 0) {
  iotjs_string_destroy(&path);
  return JS_CREATE_ERROR(COMMON, "Could not load the source.");
}
```
It will throw error when the module file is empty, I will move it, is it appropriate？

PR #476, opened 2019-01-02
module: avoid error when require an empty module fix/gh-330.

Txiaozhe:
##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/modules/iotjs_module_process.c
+++ b/src/modules/iotjs_module_process.c
@@ -146,10 +146,6 @@ JS_FUNCTION(Compile) {
 
   size_t size = 0;
   char* source = iotjs__file_read(filename, &size);
-  if (source == NULL || size == 0) {
-    iotjs_string_destroy(&path);
-    return JS_CREATE_ERROR(COMMON, "Could not load the source.");
-  }
 
   jerry_value_t jres = WrapEval(filename, strlen(filename), source, size);
   iotjs_string_destroy(&path);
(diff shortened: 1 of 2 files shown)


Txiaozhe:
#330

src/modules/iotjs_module_process.c
```diff
@@ -146,10 +146,6 @@ JS_FUNCTION(Compile) {
 
   size_t size = 0;
   char* source = iotjs__file_read(filename, &size);
-  if (source == NULL || size == 0) {
```
yorkie: I think NULL is also needed to be checked :)

yorkie:
And BTW, if you are going to fix an issue like #330, then the branch name should be fix/gh-330 :)

src/modules/iotjs_module_process.c
yorkie: See https://github.com/Rokid/ShadowNode/blob/master/src/iotjs_util.c#L30, the NULL means the file doesn't exist.

src/modules/iotjs_module_process.c
```diff
@@ -146,12 +146,11 @@ JS_FUNCTION(Compile) {
 
   size_t size = 0;
   char* source = iotjs__file_read(filename, &size);
+  jerry_value_t jres = WrapEval(filename, strlen(filename), source, size);
```
yorkie: We should verify the source before `WrapEval`, what I meant is to throw the error when source is NULL.

PR #480, opened 2019-01-10
mqtt: change the option `keepalive` in seconds.

yorkie:
The MQTT protocol uses this option in seconds, the current 60*1000
causes the mqtt keep alive is too long from server-side.

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/js/mqtt.js
+++ b/src/js/mqtt.js
@@ -40,7 +40,7 @@ function MqttClient(endpoint, options) {
     password: null,
     clientId: 'mqttjs_' + Math.random().toString(16).substr(2, 8),
     will: null,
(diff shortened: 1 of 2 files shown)


legendecas:
Does this PR resolve the dead alive issue? I did not notice any actual keep alive options changes.

lolBig:
> Does this PR resolve the dead alive issue? I did not notice any actual keep alive options changes.

since we were using milliseconds as the unit of keepalive in engine and pass it to the server, it takes a long time for the server to disconnect, @yorkie changed the keepalive timeout to seconds in the engine to keep consistence with the server

PR #445, opened 2018-12-17
module: build require.resolve

Txiaozhe:
##### Checklist

- [x] `npm test` passes
- [x] tests and/or benchmarks are included
- [x] documentation is changed or added

PR #485, opened 2019-01-16
Revert "module: add require.resolve (#445)"

yorkie:
This reverts commit 7807fe43806e86d43726e67cf7e493d65f551aa1, that causes the `require` performance problems.

/cc @legendecas @lolBig 

##### Checklist

- [x] `npm test` passes


Txiaozhe:
What scenarios can cause performance problems？

PR #489, opened 2019-02-15
n-api: factor out Node.js-agnostic N-APIs

legendecas:
Split the Node.js ECMAScript API (N-EAPI?) into its own header and
implementation files. The motivation is that the ECMAScript API stand
on its own so it might be embedded separately, implementation and all.

##### Checklist

- [x] `npm test` passes


legendecas:
Wait until `js_native_api.h` landed on Node.js LTS version.

PR #491, opened 2019-02-19
build: archive all symbols on statically linked executables

legendecas:
Fixes an issue that some unused yet exported symbols are not archived into the final executables.


yorkie:
Not yet land on 0.10.x for now, we should have a discussion with @maxbbn if the upgrading is acceptable :)

PR #492, opened 2019-02-21
v0.10.17 Release Proposal

yorkie:
Pick cbb1a4a at v0.10.x. /cc @lolBig @legendecas


legendecas:
#491 should be landed on v0.10.x too IMO.

PR #493, opened 2019-02-23
build: add platform linux-android

legendecas:
##### Notable changes:

- Merged `deps/libtuv/cmake/config` with `cmake/config`

##### Example usage

With an config file named `build-android.config` in project root:
```json
{
  "builddir": "",
  "buildlib": false,
  "buildtype": "debug",
  "clean": false,
  "config": "",
  "cmake-param": [
    "-DANDROID_ABI=armeabi-v7a",
    "-DANDROID_NDK=~/Android/sdk/ndk-bundle",
    "-DANDROID_NATIVE_API_LEVEL=23",
    "-DANDROID_TOOLCHAIN=clang"
  ],
  "profile": "./profiles/basic.profile",
  "compile-flag": [],
  "external-include-dir": [],
  "external-lib": [],
  "jerry-cmake-param": [],
  "jerry-compile-flag": [],
  "jerry-heaplimit": 4096,
  "jerry-link-flag": [],
  "jerry-lto": false,
  "jerry-memstat": false,
  "link-flag": [],
  "no-check-tidy": false,
  "no-check-test": false,
  "no-init-submodule": false,
  "no-parallel-build": false,
  "no-snapshot": false,
  "static": true,
  "sysroot": "",
  "cmake-toolchain-file": "~/Android/sdk/ndk-bundle/build/cmake/android.toolchain.cmake",
  "target-arch": "aarch64",
  "target-os": "linux-android",
  "target-board": ""
}
```

Run build script:
```sh
./tools/build.py --config=build-android.config
```

##### Checklist

- [x] preset of enabled modules on android: i.e. exclude `dbus` and `libz`
- [ ] enable CI on android
- [x] documentation of build for android


cmake/libtuv.cmake
```diff
@@ -48,6 +45,6 @@ set_property(DIRECTORY APPEND PROPERTY
 set(TUV_INCLUDE_DIR ${DEPS_TUV_SRC}/include)
 set(TUV_LIBS tuv)
 
-if("${TARGET_OS}" STREQUAL "LINUX")
+if(${TARGET_OS} STREQUAL "LINUX")
   list(APPEND TUV_LIBS pthread)
```
yorkie: Does android need this, too?

profiles/default.profile
```diff
@@ -1,2 +1,6 @@
 ENABLE_MODULE_IOTJS_BASIC_MODULES
 ENABLE_MODULE_IOTJS_CORE_MODULES
+ENABLE_MODULE_SAX
+ENABLE_MODULE_DBUS
+ENABLE_MODULE_ZLIB
```
yorkie: We could remove zlib for linux.

profiles/default.profile
```diff
@@ -1,2 +1,6 @@
...
+ENABLE_MODULE_WEBSOCKET
```
yorkie: Add MQTT, too.

yorkie:
BTW, add an android part at https://github.com/yodaos-project/ShadowNode/tree/master/docs/build.

profiles/default.profile
lolBig: should we rename the profile to linux.profile or other name? It is difficult to distinguish the meaning of basic.profile and default.profile

legendecas: IMO `default.profile` is a feature rich profile, which would enable most of features that ShadowNode supports like `websocket` and `dbus`. Rather `basic.profile` enables features that are API stable in Node.js.

cmake/libtuv.cmake
legendecas: There is no need to link pthread and rt separately on android.

Ref: https://developer.android.com/ndk/guides/stable_apis#clib

yorkie:
My fault, https://github.com/yodaos-project/ShadowNode/blob/master/docs/Getting-Started.md needs update, too.

profiles/default.profile
yorkie: `default.profile` LGTM because the `default.profile` in ShadowNode means ShadowNode itself, but that's better if we change the `basic.profile` to `nodejs.profile` :)

yorkie: And zlib should move to `nodejs.profile` or `basic.profile`, but I dunno where to disable zlib, just resolving because the zlib linking is still required though we disable the zlib module.

legendecas: AFAIK libz doesn't been linked on linux-android.

https://github.com/yodaos-project/ShadowNode/pull/493/files#diff-af3b638bc2a3e6c650974192a53c7291R119

legendecas:
> My fault, https://github.com/yodaos-project/ShadowNode/blob/master/docs/Getting-Started.md needs update, too.

Updated :)

docs/Getting-Started.md
```diff
@@ -4,6 +4,7 @@ Current supported platforms are **Linux and NuttX**
 OSX 10.10 as development host
 
 * [Build for x86 / Linux](build/Build-for-x86-Linux.md): Ubuntu 14.04 is used as a base platform.
+* [Build for Android](build/Build-for-ARM-Android.md)
```
yorkie: Android should be only for ARM?

legendecas: Actually nope. I'd remove the prefixes.

db0218:
Follow the build config, but what I built out is 32-bit Arm version, not 64-bit.

build/aarch64-linux-android/debug/bin/iotjs: ELF 32-bit LSB  shared object, ARM, EABI5 version 1 (SYSV), dynamically linked (uses shared libs), BuildID[sha1]=799e3a8d0ab40f80600bfc07ba0760e10b251219, not stripped

Could you help me ?

yorkie:
May I ask u to open an issue to file this?

PR #525, opened 2019-05-30
tuv: make loops and watchers usable after fork()

legendecas:
Added the uv_loop_fork() API that must be called in a child process to
continue using an existing loop. Internally this calls a uv__io_fork
function for each supported platform, similar to the way
uv__platform_loop_init works.

After this call, existing and new IO, async and signal watchers will
contiue working as before on all platforms, as will the
threadpool (although any threads it was using are of course gone).

On Linux and BSDs that use kqueue, existing and new fsevent watchers
will also continue to work as expected. On OS X, though, directory
fsevents will not be able to use the optimized CoreFoundation path if
they had already been used in the parent process, instead falling back
to the kqueue path used on other BSDs.

Existing fsevent watchers will not function on AIX or SunOS. This
could be relatively easily fixed by someone with AIX knowledge in the
future, but SunOS will require some additional work to keep track if
the watchers.

A new test file, test/test-fork.c, was added to contain fork-related
tests to verify functionality in the child process.

PR: libuv#846

##### Checklist

- [x] `npm test` passes


deps/libtuv/Makefile
```diff
@@ -38,7 +39,8 @@ CMAKE_DEFINES   := \
 	-DCMAKE_BUILD_TYPE=$(TUV_BUILD_TYPE) \
 	-DTARGET_PLATFORM=$(TUV_PLATFORM) \
 	-DBUILDTESTER=${TUV_BUILDTESTER} \
-	-DBUILD_HOST_HELPER=${TUV_BUILDHOSTHELPER}
+	-DBUILD_HOST_HELPER=${TUV_BUILDHOSTHELPER} \
+	-DTUV_FEATURE_PROCESS=ON
```
yorkie: Where is the define used?

deps/libtuv/src/unix/linux-core.c
```diff
@@ -121,6 +121,26 @@ int uv__platform_loop_init(uv_loop_t* loop) {
...
+  loop->backend_fd = -1;
+  uv__platform_loop_delete(loop);
+
+  err = uv__platform_loop_init(loop);
+  if (err)
+    return err;
+
+  return 0;
+  // TODO: inotify not enabled
+  // return uv__inotify_fork(loop, old_watchers);
```
yorkie: Is this comment from libuv?

legendecas: No, inotify is not enabled in libtuv.

PR #533, opened 2019-06-13
mqtt: fix mqtt msgid err

Txiaozhe:
##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/js/mqtt.js
+++ b/src/js/mqtt.js
@@ -396,11 +396,11 @@ function _getLastPacketId() {
 MqttClient.prototype.getLastPacketId = _getLastPacketId;
 
 /**
- * for Backward-compatibility
+ * @deprecated
  * @method getLastMessageId
  */
 MqttClient.prototype.getLastMessageId = util.deprecate(_getLastPacketId, 
-  'getLastMessageId is deprecated, please remove this call.');
+  'getLastMessageId is deprecated, please use `getLastPacketId` instead.');
 
 function connect(endpoint, options) {
   var client = new MqttClient(endpoint, options);
--- a/src/js/mqtt.js
+++ b/src/js/mqtt.js
@@ -276,9 +276,8 @@ MqttClient.prototype.disconnect = function(err) {
   this._socket.end();
 };
 
-function qosEqual0(qos) {
-  if (!qos || Number.isNaN(qos)) return false;
-  return qos === 0;
+function getQoS(qos) {
+  return Number.isNaN(qos) ? 0 : (qos >= 0 && qos <= 2 ? qos : 0);
 }
 
 MqttClient.prototype._nextMsgId = function() {
@@ -302,10 +301,11 @@ MqttClient.prototype.publish = function(topic, payload, options, callback) {
   if (!Buffer.isBuffer(payload)) {
     payload = new Buffer(payload);
   }
+  var qos = getQoS(options && options.qos);
   try {
     var buf = this._handle._getPublish(topic, {
-      id: qosEqual0(options && options.qos) ? 0 : this._nextMsgId(),
-      qos: (options && options.qos) || 0,
+      id: qos === 0 ? 0 : this._nextMsgId(),
+      qos: qos,
       dup: (options && options.dup) || false,
       retain: (options && options.retain) || false,
       payload: payload,
@@ -332,9 +332,10 @@ MqttClient.prototype.subscribe = function(topic, options, callback) {
     callback = callback || noop;
   }
   try {
+    var qos = getQoS(options && options.qos);
     var buf = this._handle._getSubscribe(topic, {
-      id: qosEqual0(options.qos) ? 0 : this._nextMsgId(),
-      qos: (options && options.qos) || 0,
+      id: qos === 0 ? 0 : this._nextMsgId(),
+      qos: qos,
     });
     this._write(buf, callback);
   } catch (err) {
--- a/src/js/mqtt.js
+++ b/src/js/mqtt.js
@@ -288,7 +288,7 @@ MqttClient.prototype.publish = function(topic, payload, options, callback) {
   }
   try {
     var buf = this._handle._getPublish(topic, {
-      id: this._msgId++,
+      id: ++this._msgId,
       qos: (options && options.qos) || 0,
       dup: (options && options.dup) || false,
       retain: (options && options.retain) || false,
@@ -317,7 +317,7 @@ MqttClient.prototype.subscribe = function(topic, options, callback) {
   }
   try {
     var buf = this._handle._getSubscribe(topic, {
-      id: this._msgId++,
+      id: ++this._msgId,
       qos: (options && options.qos) || 0,
     });
     this._write(buf, callback);
@@ -340,7 +340,7 @@ MqttClient.prototype.unsubscribe = function(topic, callback) {
   // TODO don't use try catch
   try {
     buf = this._handle._getUnsubscribe(topic, {
-      id: this._msgId++,
+      id: ++this._msgId,
     });
   } catch (err) {
     callback(err);
--- a/src/js/mqtt.js
+++ b/src/js/mqtt.js
@@ -277,8 +277,8 @@ MqttClient.prototype.disconnect = function(err) {
 };
 
 MqttClient.prototype._getQoS = function(qos) {
-  return  qos >= 0 && qos <= 2 ? qos : 0;
-}
+  return qos >= 0 && qos <= 2 ? qos : 0;
+};
 
 MqttClient.prototype._getNewPacketId = function() {
   if (this._packetId > MAX_MSG_ID) {
@@ -286,7 +286,7 @@ MqttClient.prototype._getNewPacketId = function() {
   }
 
   return this._packetId++;
-}
+};
 
 /**
  * @method publish
--- a/src/js/mqtt.js
+++ b/src/js/mqtt.js
@@ -274,6 +274,11 @@ MqttClient.prototype.disconnect = function(err) {
   this._socket.end();
 };
 
+function qosEqual0(qos) {
+  if (!qos || Number.isNaN(qos)) return false;
+  return qos === 0;
+}
+
 /**
  * @method publish
  * @param {String} topic
@@ -283,12 +288,16 @@ MqttClient.prototype.disconnect = function(err) {
  */
 MqttClient.prototype.publish = function(topic, payload, options, callback) {
   callback = callback || noop;
+  if (typeof options === 'function') {
+    callback = options;
+  }
+
   if (!Buffer.isBuffer(payload)) {
     payload = new Buffer(payload);
   }
   try {
     var buf = this._handle._getPublish(topic, {
-      id: ++this._msgId,
+      id: qosEqual0(options && options.qos) ? 0 : ++this._msgId,
       qos: (options && options.qos) || 0,
       dup: (options && options.dup) || false,
       retain: (options && options.retain) || false,
@@ -317,7 +326,7 @@ MqttClient.prototype.subscribe = function(topic, options, callback) {
   }
   try {
     var buf = this._handle._getSubscribe(topic, {
-      id: ++this._msgId,
+      id: qosEqual0(options.qos) ? 0 : ++this._msgId,
       qos: (options && options.qos) || 0,
     });
     this._write(buf, callback);
--- a/src/js/mqtt.js
+++ b/src/js/mqtt.js
@@ -22,7 +22,7 @@ var MQTT_PINGRESP = 13;
 var MQTT_DISCONNECT = 14;
 /* eslint-enable */
 
-var MAX_MSG_ID = 0xffff;
+var MAX_PACKET_ID = 0xffff;
 
 function noop() {}
 
@@ -281,7 +281,7 @@ MqttClient.prototype._getQoS = function(qos) {
 };
 
 MqttClient.prototype._getNewPacketId = function() {
-  if (this._packetId > MAX_MSG_ID) {
+  if (this._packetId > MAX_PACKET_ID) {
     this._packetId = 1;
   }
 
@@ -387,9 +387,9 @@ MqttClient.prototype.reconnect = function() {
 };
 
 /**
- * @method getLastMessageId
+ * @method getLastPacketId
  */
-MqttClient.prototype.getLastMessageId = function() {
+MqttClient.prototype.getLastPacketId = function() {
   return this._packetId;
 };
 
--- a/src/js/mqtt.js
+++ b/src/js/mqtt.js
@@ -22,6 +22,8 @@ var MQTT_PINGRESP = 13;
 var MQTT_DISCONNECT = 14;
 /* eslint-enable */
 
+var MAX_MSG_ID = 65535;
+
 function noop() {}
 
 /**
@@ -279,6 +281,14 @@ function qosEqual0(qos) {
   return qos === 0;
 }
 
+MqttClient.prototype._nextMsgId = function() {
+  if (this._msgId > MAX_MSG_ID) {
+    this._msgId = 1;
+  }
+
+  return this._msgId++;
+}
+
 /**
  * @method publish
  * @param {String} topic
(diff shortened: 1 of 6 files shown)


lolBig:
please provide documents for subscribe header and describe why dose `this._msgId++` not working

Txiaozhe:
The control packet of SUBSCRIBE, UNSUBSCRIBE and PUBLISH(QoS>0) must contains a nonzero packet identifier(look at [doc 2.3.1](https://mcxiaoke.gitbooks.io/mqtt-cn/content/mqtt/02-ControlPacketFormat.html))
our code(e.e.: subscribe ):
```javascript
var buf = this._handle._getSubscribe(topic, {
      id:  this._msgId++,
      qos: (options && options.qos) || 0,
});
```
The value of `this._msgId` will be `0` if we use `this._msgId++` to generate the packet identifier and `++this._msgId` is needed.

lolBig:
LGTM

yorkie:
> The control packet of SUBSCRIBE, UNSUBSCRIBE and PUBLISH(QoS>0) must contains a nonzero packet identifier(look at [doc 2.3.1](https://mcxiaoke.gitbooks.io/mqtt-cn/content/mqtt/02-ControlPacketFormat.html))
> our code(e.e.: subscribe ):
> 
> ```js
> var buf = this._handle._getSubscribe(topic, {
>       id:  this._msgId++,
>       qos: (options && options.qos) || 0,
> });
> ```
> 
> The value of `this._msgId` will be `0` if we use `this._msgId++` to generate the packet identifier and `++this._msgId` is needed.

Could you please reference the official documentation?

Txiaozhe:
Done, could please review it?

src/js/mqtt.js
```diff
@@ -274,6 +274,11 @@ MqttClient.prototype.disconnect = function(err) {
   this._socket.end();
 };
 
+function qosEqual0(qos) {
+  if (!qos || Number.isNaN(qos)) return false;
```
yorkie: What the case that qos is NaN?

src/js/mqtt.js
```diff
@@ -274,6 +274,11 @@ MqttClient.prototype.disconnect = function(err) {
...
+  return qos === 0;
+}
```
yorkie: And I'd like to add a `normalizeOpts()` function to replace this, we should use `(options && options.qos) || 0`.

src/js/mqtt.js
```diff
@@ -340,7 +349,7 @@ MqttClient.prototype.unsubscribe = function(topic, callback) {
   // TODO don't use try catch
   try {
     buf = this._handle._getUnsubscribe(topic, {
-      id: this._msgId++,
+      id: ++this._msgId,
```
yorkie: QoS is missing here?

yorkie:
> Done, could please review it?

Where is the official spec to describe this? If those are protocol-specified feature, it's better to change at MQTTPacket.

test/run_pass/test_mqtt_sub_pub.js
```diff
@@ -28,7 +28,9 @@ Promise.all([
   babeee.subscribe('u/love', common.mustCall(function() {
     console.log('babeee subscribed u/love');
     setTimeout(function() {
-      yorkie.publish('u/love', 'endless');
+      yorkie.publish('u/love', 'endless', common.mustCall(function() {
+        console.log('yorkie published u/love');
+      }));
```
yorkie: What do you wanna test by this change?

src/js/mqtt.js
Txiaozhe: when the `options` argument is undefined

yorkie:
Note that adding a related label is also required if the PR is opened by collaborators.

src/js/mqtt.js
Txiaozhe: There is not `options` argument here.

src/js/mqtt.js
```diff
@@ -283,12 +288,16 @@ MqttClient.prototype.disconnect = function(err) {
  */
 MqttClient.prototype.publish = function(topic, payload, options, callback) {
   callback = callback || noop;
+  if (typeof options === 'function') {
+    callback = options;
+  }
```
yorkie: This is unrelated, and it changes the API, revert this, please.

test/run_pass/test_mqtt_sub_pub.js
Txiaozhe: I check the case when options is undefined and the callback is the third argument, may it is not very rigorous, I will revise it.

src/js/mqtt.js
Txiaozhe: Okay.

src/js/mqtt.js
```diff
@@ -283,12 +288,16 @@ MqttClient.prototype.disconnect = function(err) {
...
+
   if (!Buffer.isBuffer(payload)) {
     payload = new Buffer(payload);
   }
   try {
     var buf = this._handle._getPublish(topic, {
-      id: this._msgId++,
+      id: qosEqual0(options && options.qos) ? 0 : ++this._msgId,
```
yorkie: ++ is not good way to represent, the root reason is that the `msgId` is starting from zero, could that start from 1 to fix this problem?

Txiaozhe: But should it be the different value when different publish?

Txiaozhe: Yeah, may ++ is not very readable:)

src/js/mqtt.js
yorkie: undefined has been covered at `!qos`

src/js/mqtt.js
yorkie: > There is not `options` argument here.

What I mean is that if the unsubscribe need qos option?

src/js/mqtt.js
```diff
@@ -317,7 +323,7 @@ MqttClient.prototype.subscribe = function(topic, options, callback) {
   }
   try {
     var buf = this._handle._getSubscribe(topic, {
-      id: this._msgId++,
+      id: qosEqual0(options.qos) ? 0 : this._msgId++,
       qos: (options && options.qos) || 0,
```
yorkie: Subscribe's qos could be removed from this option, see https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901023, only the publish's qos is used at MQTT v5.0-spec.

Txiaozhe: But our mqtt version is 3.x now, [http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718063](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718063) 
> The payload of a SUBSCRIBE packet MUST contain at least one Topic Filter / QoS pair. A SUBSCRIBE packet with no payload is a protocol violation

yorkie:
@Txiaozhe What you want to fix is documented here: https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901026.

> These MQTT Control Packets are PUBLISH (where QoS > 0), PUBACK, PUBREC, PUBREL, PUBCOMP, SUBSCRIBE, SUBACK, UNSUBSCRIBE, UNSUBACK.

As the above description, I think we only check QoS at the `.publish()`.

> Each time a Client sends a new SUBSCRIBE, UNSUBSCRIBE,or PUBLISH (where QoS > 0) MQTT Control Packet it MUST assign it a non-zero Packet Identifier that is currently unused

Firstly we should make the packet id namely `msgId` to be a non-zero Two Byte Integer, and is currently unused. Otherwise we only need do SUBSCRIBE, UNSUBSCRIBE,or PUBLISH (where QoS > 0) MQTT.

src/js/mqtt.js
Txiaozhe: there is not `qos` option in unsubscribe.

src/js/mqtt.js
```diff
@@ -274,6 +276,19 @@ MqttClient.prototype.disconnect = function(err) {
   this._socket.end();
 };
 
+function qosEqual0(qos) {
```
legendecas: `qosEqual0` would still return `false` if `qos` is `0`.

src/js/mqtt.js
```diff
@@ -22,6 +22,8 @@ var MQTT_PINGRESP = 13;
 var MQTT_DISCONNECT = 14;
 /* eslint-enable */
 
+var MAX_MSG_ID = 65535;
```
yorkie:
```suggestion
var MAX_MSG_ID = 0xffff;
```

src/js/mqtt.js
```diff
@@ -274,6 +276,18 @@ MqttClient.prototype.disconnect = function(err) {
   this._socket.end();
 };
 
+function getQoS(qos) {
+  return Number.isNaN(qos) ? 0 : (qos >= 0 && qos <= 2 ? qos : 0);
+}
+
+MqttClient.prototype._nextMsgId = function() {
+  if (this._msgId > MAX_MSG_ID) {
+    this._msgId = 1;
```
yorkie: The spec shows the packet id won't be used, that implies us to store the range of identifiers used?

src/js/mqtt.js
```diff
@@ -274,6 +276,18 @@ MqttClient.prototype.disconnect = function(err) {
   this._socket.end();
 };
 
+function getQoS(qos) {
+  return Number.isNaN(qos) ? 0 : (qos >= 0 && qos <= 2 ? qos : 0);
+}
+
+MqttClient.prototype._nextMsgId = function() {
+  if (this._msgId > MAX_MSG_ID) {
```
yorkie: How about renaming as `_packetId`?

src/js/mqtt.js
```diff
@@ -274,6 +276,18 @@ MqttClient.prototype.disconnect = function(err) {
   this._socket.end();
 };
 
+function getQoS(qos) {
+  return Number.isNaN(qos) ? 0 : (qos >= 0 && qos <= 2 ? qos : 0);
+}
+
+MqttClient.prototype._nextMsgId = function() {
```
yorkie: Suggestion: getNewPacketId()

src/js/mqtt.js
```diff
@@ -316,9 +332,10 @@ MqttClient.prototype.subscribe = function(topic, options, callback) {
     callback = callback || noop;
   }
   try {
+    var qos = getQoS(options && options.qos);
     var buf = this._handle._getSubscribe(topic, {
-      id: this._msgId++,
-      qos: (options && options.qos) || 0,
+      id: qos === 0 ? 0 : this._nextMsgId(),
```
yorkie: Subscribe seems no need to check qos.

src/js/mqtt.js
```diff
@@ -283,13 +297,15 @@ MqttClient.prototype.disconnect = function(err) {
...
+
   if (!Buffer.isBuffer(payload)) {
     payload = new Buffer(payload);
   }
+  var qos = getQoS(options && options.qos);
   try {
     var buf = this._handle._getPublish(topic, {
-      id: this._msgId++,
-      qos: (options && options.qos) || 0,
+      id: qos === 0 ? 0 : this._nextMsgId(),
```
yorkie: The id should be undefined(unset) if QoS is zero.

src/js/mqtt.js
Txiaozhe: Now we have not check the ack id, so it is not very emergency now. Next we need perfect it.

yorkie: Ok, seems good to me now.

src/js/mqtt.js
```diff
@@ -274,6 +276,18 @@ MqttClient.prototype.disconnect = function(err) {
   this._socket.end();
 };
 
+MqttClient.prototype._getQoS = function(qos) {
+  return Number.isNaN(qos) ? 0 : (qos >= 0 && qos <= 2 ? qos : 0);
```
legendecas: No need to check if qos is NaN, `(NaN >= 0 && NaN <= 2) === false`.

src/js/mqtt.js
```diff
@@ -22,6 +22,8 @@ var MQTT_PINGRESP = 13;
 var MQTT_DISCONNECT = 14;
 /* eslint-enable */
 
+var MAX_MSG_ID = 0xffff;
```
yorkie:
```suggestion
var MAX_PACKET_ID = 0xffff;
```

src/js/mqtt.js
```diff
@@ -274,6 +276,18 @@ MqttClient.prototype.disconnect = function(err) {
   this._socket.end();
 };
 
+MqttClient.prototype._getQoS = function(qos) {
```
yorkie:
```suggestion
MqttClient.prototype._getQoS = function getQoS(val) {
```

src/js/mqtt.js
```diff
@@ -283,13 +297,15 @@ MqttClient.prototype.disconnect = function(err) {
...
+
   if (!Buffer.isBuffer(payload)) {
     payload = new Buffer(payload);
   }
+  var qos = this._getQoS(options && options.qos);
   try {
     var buf = this._handle._getPublish(topic, {
-      id: this._msgId++,
-      qos: (options && options.qos) || 0,
+      id: qos === 0 ? 0 : this._getNewPacketId(),
```
yorkie:
```suggestion
      id: qos === 0 ? undefined : this._getNewPacketId(),
```

test/run_pass/test_mqtt_qos.js
```diff
@@ -0,0 +1,36 @@
...
+  }, (err) => {
+    assert.strictEqual(babeee._getQoS(1), 1);
+  });
+
+  babeee.subscribe('u/love3', {
+    qos: 5
+  }, (err) => {
+    assert.strictEqual(babeee._getQoS(0), 0);
+  });
+});
```
yorkie: Should assert on packetId, too.

src/js/mqtt.js
```diff
@@ -373,7 +390,7 @@ MqttClient.prototype.reconnect = function() {
  * @method getLastMessageId
  */
 MqttClient.prototype.getLastMessageId = function() {
```
yorkie: Change to `getLastPacketId`, and deprecates `getLastMessageId`?

src/js/mqtt.js
Txiaozhe: Can packet id be undefined?

yorkie: At MQTT spec, the behavior needs to be unset if QoS is zero.

src/js/mqtt.js
```diff
@@ -370,10 +387,10 @@ MqttClient.prototype.reconnect = function() {
 };
 
 /**
- * @method getLastMessageId
+ * @method getLastPacketId
  */
-MqttClient.prototype.getLastMessageId = function() {
-  return this._msgId;
+MqttClient.prototype.getLastPacketId = function() {
```
yorkie: Backward-compatibility is required.

yorkie:
@Txiaozhe You should add your tests to https://github.com/yodaos-project/ShadowNode/blob/master/test/testsets.json.

src/js/mqtt.js
```diff
@@ -369,12 +386,21 @@ MqttClient.prototype.reconnect = function() {
...
+
+/**
+ * for Backward-compatibility
  * @method getLastMessageId
  */
-MqttClient.prototype.getLastMessageId = function() {
-  return this._msgId;
-};
+MqttClient.prototype.getLastMessageId = util.deprecate(_getLastPacketId, 
+  'getLastMessageId is deprecated, please remove this call.');
```
yorkie:
```suggestion
  'getLastMessageId is deprecated, please use `getLastPacketId` instead.');
```

src/js/mqtt.js
```diff
@@ -369,12 +386,21 @@ MqttClient.prototype.reconnect = function() {
...
+};
+
 /**
+ * @method getLastPacketId
+ */
+MqttClient.prototype.getLastPacketId = _getLastPacketId;
+
+/**
+ * for Backward-compatibility
  * @method getLastMessageId
```
yorkie: add @deprecated is good.

src/js/mqtt.js
Txiaozhe:
```c
if (qos > 0)
	writeInt(&ptr, packetid);
```
if qos == 0, it will not set the packetid, so 0 is suitable.

PR #544, opened 2019-07-13
MQTT: fix connect timeout not working

lolBig:
##### Checklist

- [x] `npm test` passes


src/js/mqtt.js
```diff
@@ -105,10 +110,19 @@ MqttClient.prototype.connect = function() {
...
+};
+
 /**
  * @method _onconnect
  */
 MqttClient.prototype._onconnect = function() {
+  if (this._connectTimer) {
+    clearTimeout(this._connectTimer);
+    this._connectTimer = null;
+  }
```
yorkie: See https://github.com/mqttjs/MQTT.js#mqttclientstreambuilder-options, we should reset the state when receiving the CONNACK packet.

PR #547, opened 2019-07-24
MQTT: fix event flow & add more tests

lolBig:
##### Checklist

- [x] `npm test` passes

### DIFF
--- a/docs/api/MQTT.md
+++ b/docs/api/MQTT.md
@@ -1,9 +1,9 @@
 # MQTT
 
-This module is a native module for the `MQTT` protocol, which is compatible with 
+This module is a native module for the `MQTT` protocol, which is compatible with
 the [MQTT.js](https://github.com/mqttjs/MQTT.js).
 
-**Example**
+## Example
 
 ```js
 var mqtt = require('mqtt');
@@ -17,7 +17,7 @@ client.on('connect', function () {
 client.on('message', function (topic, message) {
   // message is Buffer
   console.log(message.toString());
-  client.end();
+  client.disconnect();
 });
 ```
 
@@ -53,7 +53,7 @@ Emitted when a reconnect starts.
 
 ### Event `'offline'`
 
-Emitted when the client goes offline.
+Emitted when the client from online goes offline.
 
 ### Event `'error'`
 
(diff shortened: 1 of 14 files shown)


lolBig:
issue: #548

docs/api/MQTT.md
```diff
@@ -17,7 +17,7 @@ client.on('connect', function () {
 client.on('message', function (topic, message) {
   // message is Buffer
   console.log(message.toString());
-  client.end();
+  client.disconnect();
```
yorkie: The MQTT API keeps consistent with [MQTT.js](https://github.com/mqttjs/MQTT.js), which doesn't use `disconnect()`.

docs/api/MQTT.md
```diff
@@ -111,10 +111,13 @@ Unsubscribe from a topic.
 * `force` {Boolean} passing it to true will close the client right away.
 * `callback` {Function} will be called when the client is closed. This parameter is optional.
 
-Close the client.
+### client.disconnect([err, [, callback]])
```
yorkie: Ditto

docs/api/MQTT.md
```diff
@@ -111,10 +111,13 @@ Unsubscribe from a topic.
...
-### client.reconnect()
+* `err` {Error} throw an error before disconnect. This parameter is optional.
+* `callback` {Function} will be called when the client is closed. This parameter is optional.
 
-Connect again using the same options as connect().
+Disconnect the client right away The `client` can safety be unref after call this method.
+
+### client.reconnect()
 
-For more detailed, please see [MQTT.js Readme](https://github.com/mqttjs/MQTT.js#client).
```
yorkie: As I have commented, this rule should not be removed from this documentation.

yorkie:
BTW, if a PR branch is to fix a specific issue, a suggestion name shall be like `fix/gh-648`.

docs/api/MQTT.md
```diff
@@ -113,8 +113,17 @@ Unsubscribe from a topic.
 
 Close the client.
 
+### client.disconnect([err, [, callback]])
```
legendecas: What's the difference between `client.end` and `client.disconnect`?

lolBig: they both do the same thing, the MQTT.js use `end` and we use `disconnect` before, so keep `disconnect` and add `end` to make compatible with `MQTT.js` and old apps

PR #562, opened 2019-09-21
constants: refactors the module to use os defines

yorkie:
It fixes https://github.com/yodaos-project/ShadowNode/issues/560 by fixing the missing signal `SIGCHLD`.

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/modules/iotjs_module_constants.c
+++ b/src/modules/iotjs_module_constants.c
@@ -17,7 +17,7 @@
 #include "iotjs_module.h"
 #include "mbedtls/ssl.h"
 
-#define SET_CONSTANT(object, constant)                           \
+#define NODE_DEFINE_CONSTANT(object, constant)                   \
   do {                                                           \
     iotjs_jval_set_property_number(object, #constant, constant); \
   } while (0)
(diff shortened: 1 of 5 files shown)


src/modules/iotjs_module_constants.c
```diff
@@ -30,33 +30,192 @@
...
-  SET_CONSTANT(constants, S_IFREG);
-  SET_CONSTANT(constants, S_IFLNK);
-  SET_CONSTANT(constants, S_IFSOCK);
-  SET_CONSTANT(constants, TLS_CHUNK_MAX_SIZE);
-  SET_CONSTANT(constants, INT32_MAX_VALUE);
-  // uv constants
-  SET_CONSTANT(constants, UV_PRIORITY_LOW);
-  SET_CONSTANT(constants, UV_PRIORITY_HIGHEST);
+  jerry_value_t exports = jerry_create_object();
+  jerry_value_t os = jerry_create_object();
```
legendecas: Could this variable be moved to its nearest usage line 217?

yorkie: It makes the first-level objects more clear, we might have more constant groups, if we were to move there as the Latest Usage Rule, that looks like the below:

```js
var exports = {};
//...
var os = {};
exports.os = os;
var signals = {};
os.signals = signals;
//...
var platforms = {};
os.platforms = platforms;

var uv = {};
exports.uv = uv;
var errnos = {};
uv.errnos = errnos;
//...
```
And declare first-level objects firstly looks like:
```js
var exports = {};
var os = {};
var uv = {};
exports.os = os;
exports.uv = uv;
//...
var signals = {};
os.signals = signals;
var platforms = {};
os.platforms = platforms;
//...

var errnos = {};
uv.errnos = errnos;
//...
```
The latter looks more comfortable to me :p

yorkie:
Updated the testset and add SIGCHLD into our old simple signal case, too.

PR #563, opened 2019-09-21
process: use uv's exception if starting an invalid signal handler

yorkie:
This keeps consistent with Node.js' behavior, it uses uv's exception instead of ignoring for this case.

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/js/iotjs.js
+++ b/src/js/iotjs.js
@@ -522,10 +522,6 @@
       if (!isSignal(type) || signalWraps[type] !== undefined) {
         return;
       }
-      if (type === 'SIGKILL' || type === 'SIGSTOP') {
-        // see sigaction(2), SIGKILL/SIGSTOP are not supported, just skip it.
-        return;
-      }
       var Signal = Module.require('signal');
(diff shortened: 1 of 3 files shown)


yorkie:
> AssertionError: Missing expected exception: should throw uv_signal_start(SIGSTOP) invalid args

It seems that the macOS on TravisCI does allow handling `SIGSTOP`(it's no problems on my macbook), what's the idea about this failure, guys? @legendecas @lolBig

legendecas:
AFAIK, SIGKILL and SIGSTOP could not be caught.

Refs: http://man7.org/linux/man-pages/man7/signal.7.html

legendecas:
The case also failed locally on my macOS mojave 10.14.6.

legendecas:
macOS has a different SIGSTOP value (17) than linux's (19), while 19 on macOS is SIGCONT which could safely be caught.

Since then, #562 would be blocking this PR from continuing its work.

yorkie:
Thanks @legendecas, the CI seems green :)

PR #565, opened 2019-09-24
stream: use empty function if no callback with `write()`

yorkie:
Add this check https://github.com/nodejs/node/blob/master/lib/_stream_writable.js#L299-L300 at ShadowNode, to reproduce the problem, try the following script:

```js
require('fs').createWriteStream('./test').write('foobar')
```

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/js/stream_writable.js
+++ b/src/js/stream_writable.js
@@ -20,7 +20,7 @@ var util = require('util');
 var Stream = stream.Stream;
 var defaultHighWaterMark = 128;
 
-function nop() {}
+function noop() {}
 
 function WriteReq(chunk, encoding, callback) {
   this.chunk = chunk;
@@ -99,7 +99,7 @@ Writable.prototype.write = function(chunk, encoding, callback) {
   }
 
   if (typeof callback !== 'function')
-    callback = nop;
+    callback = noop;
 
   var state = this._writableState;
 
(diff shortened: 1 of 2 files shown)


src/js/stream_writable.js
```diff
@@ -96,6 +98,9 @@ Writable.prototype.write = function(chunk, encoding, callback) {
     encoding = null;
   }
 
+  if (typeof callback !== 'function')
```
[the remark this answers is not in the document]
yorkie: See https://github.com/nodejs/node/blob/master/lib/_stream_writable.js#L299-L300, this patch only follows the Node.js' behavior, if we do this change, it should be landed on Node.js firstly?

src/js/fs.js
```diff
@@ -857,9 +857,16 @@ WriteStream.prototype._write = function(data, encoding, cb) {
       if (this.autoClose) {
         this._destroy();
       }
-      return cb(er);
+      process.nextTick(onwrite);
```
legendecas: Is there any reason for the callback being deferred to next tick?

src/js/stream_writable.js
```diff
@@ -20,6 +20,8 @@ var util = require('util');
 var Stream = stream.Stream;
 var defaultHighWaterMark = 128;
 
+function nop() {}
```
legendecas: Might be `noop`?

PR #569, opened 2019-10-17
os: add method `tmpdir()`

yorkie:
Implements the Node.js `os` API: https://nodejs.org/dist/latest-v12.x/docs/api/os.html#os_os_tmpdir

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/docs/api/OS.md
+++ b/docs/api/OS.md
@@ -43,7 +43,7 @@ The properties available on the assigned network address object include:
 * `scopeid` {Number} The numeric IPv6 scope ID (only specified when family is IPv6)
 * `cidr` {String} The assigned IPv4 or IPv6 address with the routing prefix in CIDR notation. If the netmask is invalid, this property is set to null
 
-```js
+```
 {
   lo: [
     {
@@ -110,3 +110,9 @@ of the current process is used.
 
 The `priority` input must be an integer between `-20` (high priority) and `19`
 (low priority).
+
+## os.tmpdir()
+
+* Returns: {string}
+
+The `os.tmpdir()` method returns a string specifying the operating system's default directory for temporary files.
--- a/src/js/os.js
+++ b/src/js/os.js
@@ -88,3 +88,18 @@ Object.defineProperty(exports, 'EOL', {
     return '\n';
   }
 });
+
+function isStringEndsWith(str, search) {
+  var len = str.length;
+  return str.substring(len - search.length, len) === search;
+}
+
+exports.tmpdir = function tmpdir() {
+  var path = process.env.TMPDIR ||
+             process.env.TMP ||
+             process.env.TEMP || '/tmp';
+  if (path.length > 1 && isStringEndsWith(path, '/'))
+    path = path.slice(0, -1);
+
+  return path;
+};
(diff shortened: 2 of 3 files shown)


yorkie:
Note: this is a prepared PR for adding the `fsync` and `fdatasync()` :)

src/js/os.js
```diff
@@ -88,3 +88,18 @@ Object.defineProperty(exports, 'EOL', {
...
   }
 });
+
+function isStringEndsWith(str, search) {
+  var len = str.length;
+  return str.substring(len - search.length, len) === search;
+}
+
+exports.tmpdir = function tmpdir() {
+  var path = process.env.TMPDIR ||
```
lolBig: why not get env from [getenv(3)](http://man7.org/linux/man-pages/man3/getenv.3.html)

yorkie: What's the benefits in that way? Actually we did set the env at [`GetEnvironArray`](https://github.com/yodaos-project/ShadowNode/blob/master/src/modules/iotjs_module_process.c#L423) via [`environ(7)`](http://man7.org/linux/man-pages/man7/environ.7.html).

yorkie:
Restarts the failed job(https://travis-ci.com/yodaos-project/ShadowNode/jobs/246598500) caused by networks.

test/run_pass/test_os.js
```diff
@@ -35,3 +35,21 @@ var expected = [
 ];
 
 assert.deepStrictEqual(actual, expected);
+
+// test for tmpdir
```
legendecas: A default case shall be covered. i.e. no `TMPDIR` nor `TMP` and `TEMP` environs are set.

src/js/os.js
lolBig: in Windows, Node.js is also in this way，but on other [platforms](https://github.com/nodejs/node/blob/908292cf1f551c614a733d858528ffb13fb3a524/lib/os.js#L133) are via [getenv](https://github.com/nodejs/node/blob/a8277ddb5505056e3d66d02051c9195310c5c715/src/node_credentials.cc#L61), i am not sure why Node.js is doing this.

legendecas: getenv is only used if node environment has not been setup.

docs/api/OS.md
```diff
@@ -43,7 +43,7 @@ The properties available on the assigned network address object include:
 * `scopeid` {Number} The numeric IPv6 scope ID (only specified when family is IPv6)
 * `cidr` {String} The assigned IPv4 or IPv6 address with the routing prefix in CIDR notation. If the netmask is invalid, this property is set to null
 
-```js
```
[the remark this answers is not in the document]
yorkie: Yea, the below lines actually are not in JavaScript.

src/js/os.js
yorkie: See https://github.com/nodejs/node/pull/18511 for all related history, this seems a historic design problem that allows packages sharing via `process.env`(write/read), but anyway, we have to handle this security, thanks @lolBig

test/run_pass/test_os.js
yorkie: Ok, BTW I just copied these tests from here: https://github.com/nodejs/node/blob/master/test/parallel/test-os.js#L46-L78.

